PR #50, opened 2014-08-30
add hide_identity and hide_version parameters, and update the

buzzdeee:
template to add those parameters to the server: configuration.

default for both variables is true, since there should not really be a need to give that information out to the clients if they ask, the administrator should know what DNS server is running ;)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -42,7 +42,9 @@ class unbound (
   $num_queries_per_thread = undef,
   $outgoing_range         = undef,
   $so_rcvbuf              = undef,
-  $tcp_upstream           = false
+  $tcp_upstream           = false,
+  $hide_identity          = true,
+  $hide_version           = true,
 ) inherits unbound::params {
 
   package { $package_name:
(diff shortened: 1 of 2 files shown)


gbjtv:
Looks like it got reverted a commit later? I don't see this in master anymore

xaque208:
Looks there to me.

gbjtv:
https://github.com/xaque208/puppet-unbound/commit/df87e30632ecc1dded7dbd9e1acd00dac16649bc

Was removed here.

xaque208:
A fresh clone and running `git branch --contains df87e30632ecc1dded7dbd9e1acd00dac16649bc` yields an indication that referenced commit is in the master branch.  What am I missing?  There was a messed up commit a while ago that was pretty odd, but I believe that to be resolved.

gbjtv:
In df87e30632ecc1dded7dbd9e1acd00dac16649bc, file manifests/init.pp , line number 46 + 47. The $hide_identity and $hide_version variables were removed and never replaced with $unbound::params:: equivalents.

xaque208:
You're right about the missing params from that commit.  That looks to be something that I just overlooked when merging rather than the commit itself missing.  Care to submit a PR to add them back in?

PR #66, opened 2014-11-28
Run unbound-control-setup, in order to create the certificates

buzzdeee:
for unbound-control. Seems to be required with unbound > 1.5

I updated one of my DNS servers today, and found unbound not starting up anymore automatically.
reason seems to be, that unbound-control now defaults to SSL secure communication.
However, its fairly hardcoded.

Do you think, unbound::remote should inherit unbound::params, and take the
group name, and $configdir from there?

### DIFF
--- a/manifests/remote.pp
+++ b/manifests/remote.pp
@@ -21,4 +21,9 @@ class unbound::remote (
     target  => $config_file,
     content => template('unbound/remote.erb'),
   }
+
+  exec { 'unbound-control-setup':
+    command => 'unbound-control-setup -d /var/unbound/etc && chgrp _unbound /var/unbound/etc/unbound_*',
+    creates => '/var/unbound/etc/unbound_server.pem',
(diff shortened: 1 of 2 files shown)


manifests/remote.pp
```diff
@@ -21,4 +23,13 @@
     target  => $config_file,
     content => template('unbound/remote.erb'),
   }
+
+  exec { 'unbound-control-setup':
+    command => "unbound-control-setup -d ${confdir} && \
+		chgrp ${group} ${server_key_file} \
```
xaque208: I'd prefer that these `chgrp` be broken out into file resources for readability if thats not too much trouble.

buzzdeee:
indeed, a file resouce, managing permissions and mode of the files looks much cleaner.
How about this update then.

xaque208:
Excellent, thank you.

PR #67, opened 2014-12-01
Updates to trust anchoring

b4ldr:
The unbound process needs write access to the directory where the trust anchor file exists as it writes the new file in the same dir and then preforms a move operation to ensure the update is atomic.  If the process does not have write permission you will see an error like the following

Dec  1 10:38:39 host unbound: [11768:0] error: could not open autotrust file for writing, /etc/unbound/root.anchor.11768-0: Permission denied

This merge does the following 
   * adds a new directory which is writable and stores the anchor file.  I considered changing the permissions of confdir however it felt as if it was downgrading the security of other files.
   * moves the anchor file to this new directory
   * initiate this file using unbound-anchor instead of a hardcoded DS record.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,7 +4,9 @@
 #
 class unbound (
   $access                       = $unbound::params::access,
-  $anchor_file                  = $unbound::params::anchor_file,
+  $anchor_fetch_command         = $unbound::params::anchor_fetch_command,
+  $anchor_file                  = $unbound::params::auto_trust_anchor_file,
+  $auto_trust_anchor_file       = $unbound::params::auto_trust_anchor_file,
   $chroot                       = $unbound::params::chroot,
   $conf_d                       = $unbound::params::conf_d,
   $confdir                      = $unbound::params::confdir,
@@ -49,6 +51,7 @@ class unbound (
   $prefetch_key                 = $unbound::params::prefetch_key,
   $private_domain               = $unbound::params::private_domain,
   $root_hints_url               = $unbound::params::root_hints_url,
+  $runtime_dir                  = $unbound::params::runtime_dir,
   $rrset_cache_size             = $unbound::params::rrset_cache_size,
   $rrset_cache_slabs            = $unbound::params::rrset_cache_slabs,
   $service_name                 = $unbound::params::service_name,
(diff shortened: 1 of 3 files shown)


xaque208:
Reviewing now.  I've been neglecting my github duties, apologies.

xaque208:
Just for the sake of understanding, why not just set the owner on the confdir?  It doesn't look like we manage the permissions on that directory, but quite likely should be owned by the unbound user for the reason you raise in the OP.

b4ldr:
I didn't want to change the ownership of the config directory because if the unbound process was exploited one would be able to rewrite the unbound config file.  Admittedly if they had exploited the unbound process they would be able to modify this in memory however that is much harder to do then just getting a shell and modifying the config file and it wouldn't survive a restart.  Im probably been to causions but wanted to add it to a seperate directory incase you had intended the permissions to be restrictive

xaque208:
I'd rather stick to whatever unbound defaults to and let the admin do what is needed elsewhere.  Its been long enough of using this module, I don't actually know what the default is.  I just don't want to impose too many world views on this module if we can avoid it.

b4ldr:
By default unbound doesn't initialise the auto-trust-anchor-file file, however i would argue that using the unbound-anchor tool is more inline with what nlnetlabs would expect[1].  

As to the location there is no  unbound needs to write to it and the directory it resides however it doesn't need to write to any of the other configs.  I think unbound defaults to the config dir for this is however looking at Debian,  the file goes into /var/lib/unbound/ which is owned by unbound.  I guess the best way foward would be default to  
  $runtime_dir  = $confdir 

unless we know that the platform uses a custom path.  e.g. for debian  
   $runtime_dir  = /var/lib/unbound

if you agree i'll rework this

[1]https://www.unbound.net/documentation/howto_anchor.html

igalic:
agreed!

manifests/init.pp
```diff
@@ -4,7 +4,9 @@
 #
 class unbound (
   $access                       = $unbound::params::access,
-  $anchor_file                  = $unbound::params::anchor_file,
+  $anchor_fetch_command         = $unbound::params::anchor_fetch_command,
+  $anchor_file                  = $unbound::params::auto_trust_anchor_file,
```
xaque208: Can this line be removed?

b4ldr:
The fetch command is needed.  I added the following to make it backwards compatible with older releases as i renamed that variable to make it consistent with the config file naming convention. 

```puppet
 $anchor_file                  = $unbound::params::auto_trust_anchor_file,
```

Sorry for the late update been ill will be a bit more on the ball this week

igalic:
no worries! get well, @b4ldr!

PR #68, opened 2015-01-13
create directory before exec

f3rr:
create directory before exec, otherwise exec will fail because of missing directory


xaque208:
Cheers, thanks.

PR #69, opened 2015-02-07
Exec calls should have a full path.

robbat2:
Error: Failed to apply catalog: Validation of Exec[unbound-control-setup]
failed: 'unbound-control-setup -d /etc/unbound' is not qualified and no path
was specified. Please qualify the command or specify a path. at
.../modules/unbound/manifests/remote.pp:25


robbat2:
I'm going to reopen with just this change.

PR #70, opened 2015-02-07
unbound-checkconf: validate new conf before use

robbat2:
I managed to quietly break my unbound because of the merge order of
stubs; so add some validation now.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -68,6 +68,8 @@ class unbound (
   $val_permissive_mode          = $unbound::params::val_permissive_mode,
   $verbosity                    = $unbound::params::verbosity,
   $custom_server_conf           = $unbound::params::custom_server_conf,
+  $checkconf_enable             = $unbound::params::checkconf_enable,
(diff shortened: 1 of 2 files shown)


igalic:
i like the general idea, but we still have a global ordering [that says](https://github.com/xaque208/puppet-unbound/blob/master/manifests/init.pp#L133-L135):
```puppet
concat { $config_file:
  notify => Service[$service_name],
}
```
***
i'd prefer those two commits to be merged.

robbat2:
Squashed as requested.

If we wanted to require the checkconf passes, we could also add Exec['checkconf'] to the service require if $checkconf_enable is true.

xaque208:
If we made this check as default, we would be able to get rid of the notify that igalic pointed to.  I'ld almost be in favor of that, since it would mean that we check the state of the config before every service reload, which I think I'd prefer as a user anyway.

xaque208:
Closing in favor of https://github.com/xaque208/puppet-unbound/pull/84

PR #71, opened 2015-02-07
Ensure local-zone is always under server.

robbat2:
If you create resources with stub before local-zone, the local-zone
entries end up in the stubs, causing a validation failure.

To solve this, force the stubs to come later in the configuration file.

Example code to trigger the bug (with the hashes non-empty):
create_resources(unbound::stub, hiera_hash('unbound::zones::stub', {}), {})
create_resources(unbound::local_zone, hiera_hash('unbound::zones::local', {}), {})

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -12,7 +12,7 @@ define unbound::stub (
   $config_file = $unbound::params::config_file
 
   concat::fragment { "unbound-stub-${name}":
-    order   => '05',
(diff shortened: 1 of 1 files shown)


igalic:
i wish concat fragment had a `before` that you could use on tags, or title matches…

xaque208:
@igalic That would be a nice feature of concat.

I'm +1 on this, but I've not validated the issue.

PR #72, opened 2015-02-07
Exec calls should have a full path.

robbat2:
Error: Failed to apply catalog: Validation of Exec[unbound-control-setup]
failed: 'unbound-control-setup -d /etc/unbound' is not qualified and no path
was specified. Please qualify the command or specify a path. at
.../modules/unbound/manifests/remote.pp:25

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -91,6 +91,7 @@ class unbound::params {
   $conf_d                     = "${confdir}/conf.d"
   $config_file                = "${confdir}/unbound.conf"
   $control_enable             = false
+  $control_setup_path         = '/usr/sbin/unbound-control-setup'
   $directory                  = $confdir
   $dlv_anchor_file            = undef
   $do_ip4                     = true
--- a/manifests/remote.pp
+++ b/manifests/remote.pp
@@ -23,7 +23,7 @@ class unbound::remote (
   }
 
   exec { 'unbound-control-setup':
-    command => "unbound-control-setup -d ${confdir}",
+    command => "${unbound::params::control_setup_path} -d ${confdir}",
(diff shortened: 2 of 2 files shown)


manifests/params.pp
```diff
@@ -91,6 +91,7 @@
   $conf_d                     = "${confdir}/conf.d"
   $config_file                = "${confdir}/unbound.conf"
   $control_enable             = false
+  $control_setup_path         = '/usr/sbin/unbound-control-setup'
```
igalic: i'm pretty and sure that this will be different on different platforms…

maybe we should just use PATH in the exec?

robbat2: It needs to be fully specified, otherwise the catalog does not apply. If different platforms have it elsewhere, just move it in params to select based on OS/etc or override with hiera.

xaque208: I think what @igalic is suggesting, is that do avoid having to add the full path in params, is to add a `path =>` paramater to the `exec` resource.  This way when the call is made to the command, puppet will look in the path for the binary.

igalic: exactly…
and that way it can easily be expanded to new platforms with different paths.

xaque208:
FWIW, I usually place an Exec {} in site.pp to include `path` so these kinds of things don't happen.  If we can address the hardcode for other platforms, I like this approach.  Perhaps an update to include path as a param on the exec calls would be better here.  I'm open to ideas here.

xaque208:
Well, its /usr/sbin/unbound-control-setup on both OpenBSD and FreeBSD, as they both ship with core.

igalic:
ohkay then

PR #74, opened 2015-02-14
Begin spec testing

xaque208:
This just ensures that the catalog compiles on all supported platforms
and lays the groundwork for future rspec-puppet.


.gitignore
```diff
@@ -2,7 +2,6 @@
 Gemfile.lock
 .*.sw?
 pkg
-spec/fixtures
```
igalic: why remove this from gitignore?

spec/spec_helper.rb
```diff
@@ -0,0 +1,8 @@
+require 'rspec-puppet'
```
igalic: oh, you're not using puppetlabs-rspec-helper here?

xaque208:
@igalic I just wanted something basic for the time being.  Feel free to make improvements where warranted.

.gitignore
xaque208: I actually think I removed that only for temporary addition of the modules link.  That can likely be added back in.

igalic:
*nod*

PR #75, opened 2015-02-25
Use str2bool so always-string data from hiera can still be used as if it...

rswarts:
... were a boolean

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -17,7 +17,7 @@ define unbound::stub (
     content => template('unbound/stub.erb'),
   }
 
-  if $insecure == true {
+  if str2bool($insecure) == true {
     concat::fragment { "unbound-stub-${name}-insecure":
       order   => '01',
       target  => $config_file,


manifests/stub.pp
```diff
@@ -17,7 +17,7 @@
     content => template('unbound/stub.erb'),
   }
 
-  if $insecure == true {
+  if str2bool($insecure) == true {
```
xaque208: If I remember correctly, this will fail if $insecure is a string.

rswarts: > str2bool: This converts a string to a boolean. This attempts to convert strings that contain values such as '1', 't', 'y', and 'yes' to 'true' and strings that contain values such as '0', 'f', 'n', and 'no' to 'false'. Type: rvalue

It will also leave things that are already booleans intact. I have tested this on the strings "true" and "false" and it works fine.

xaque208: Okay, I must be remembering an older version of the module.

xaque208:
I'm +1 on this change, but could you tidy up the commit a bit so the first line is under 80 characters?

rswarts:
Of course, will this do?

igalic:
yes! thanks @rswarts!

PR #77, opened 2015-03-15
Make it work with strict variables

### LINKED ISSUE #76: Make it strict_variables-safe
```Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Undefined variable "unbound::params::group"; Undefined variable "group" at /etc/puppet/environments/billing/modules/unbound/manifests/init.pp:20 on node ip-172-20-29-111.ec2.internal
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run```

Pull request probably incoming shortly.

robinbowes:
Make sure all vars in params.pp are assigned values (even if just to undef) to the module works under strict_variables=true.

Fixes #76

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -5,7 +5,6 @@
 class unbound (
   $access                       = $unbound::params::access,
   $anchor_fetch_command         = $unbound::params::anchor_fetch_command,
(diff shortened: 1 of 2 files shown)


igalic:
thank you @robinbowes!

PR #78, opened 2015-03-15
Set correct runtime dir

robinbowes:
On RHEL, runtime dir is ```/var/lib/unbound```


igalic:
pretty cool, after merging #77, this one's now unmergable :(

igalic:
merged this manually

PR #79, opened 2015-03-15
RedHat does not install wget by default, but curl is available

robinbowes:
wget is not installed by default on a minimal RHEL install. Use curl (which is) instead.


igalic:
merged this one manually, too

PR #80, opened 2015-03-23
Update Rakefile from skeleton

xaque208:
This is simply to keep up to date with testing.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,8 +1,3 @@
-require 'rspec-puppet'
-
-fixture_path = File.expand_path(File.join(__FILE__, '..', 'fixtures'))
-
-RSpec.configure do |c|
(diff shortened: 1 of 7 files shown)


xaque208:
It looks like concat is causing the failure due to the test not setting the fact for osfamily.

spec/spec_helper.rb
```diff
@@ -1,8 +1,3 @@
...
-
-fixture_path = File.expand_path(File.join(__FILE__, '..', 'fixtures'))
-
-RSpec.configure do |c|
-  c.module_path = File.join(fixture_path, 'modules')
-  c.manifest_dir = File.join(fixture_path, 'manifests')
-end
+require 'puppetlabs_spec_helper/module_spec_helper'
+require 'rspec-puppet-facts'
+include RspecPuppetFacts
```
igalic: you could set the facts concat requires here.

PR #81, opened 2015-03-23
Use the master branches status as build indicator

xaque208:
Without this, the build indicator will flap on the latest build, which
may not be an accurate reflection of the code.  This is to keep the
build status accurate according to the master branch.


igalic:
+1

PR #83, opened 2015-04-14
Removed unnecessary references to params to avoid inheritance issues

t0mmyt:
Pretty much that.  None of the sub classes should be called directly so $unbound:: is already in scope.  By calling or referring to unbound::params values defined when calling the unbound class are forgotten about. (e.g. an alternative $config_file).

### DIFF
--- a/manifests/forward.pp
+++ b/manifests/forward.pp
@@ -7,9 +7,7 @@ define unbound::forward (
   $zone = $name,
 ) {
 
-  include unbound::params
-
-  $config_file = $unbound::params::config_file
+  $config_file = $unbound::config_file
 
   concat::fragment { "unbound-forward-${name}":
     order   => '20',
--- a/manifests/remote.pp
+++ b/manifests/remote.pp
@@ -6,15 +6,15 @@ class unbound::remote (
   $enable            = true,
   $interface         = ['::1', '127.0.0.1'],
   $port              = 953,
-  $server_key_file   = "${unbound::params::confdir}/unbound_server.key",
-  $server_cert_file  = "${unbound::params::confdir}/unbound_server.pem",
-  $control_key_file  = "${$unbound::params::confdir}/unbound_control.key",
-  $control_cert_file = "${$unbound::params::confdir}/unbound_control.pem",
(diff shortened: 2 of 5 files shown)


manifests/remote.pp
```diff
@@ -6,15 +6,15 @@
...
-  $group             = $unbound::params::group,
-  $confdir           = $unbound::params::confdir,
-) inherits unbound::params {
+  $server_key_file   = "${unbound::confdir}/unbound_server.key",
+  $server_cert_file  = "${unbound::confdir}/unbound_server.pem",
+  $control_key_file  = "${unbound::confdir}/unbound_control.key",
+  $control_cert_file = "${unbound::confdir}/unbound_control.pem",
+  $group             = $unbound::group,
+  $confdir           = $unbound::confdir,
+) inherits unbound {
```
igalic: i'm a bit unhappy about **this** change.
normally, sub-classes should inherit from *only* from ::params, and only if they are top-level classes.
otherwise, they shouldn't have no parameter themselves (and be called from init.pp)

igalic:
the rest looks good imo, :+1:

manifests/remote.pp
t0mmyt: That makes sense, I think I got a little trigger happy on my search and replace.

xaque208: I'd agree with @igalic here.  Inheritance isn't really recommended, and just about the only place its deemed acceptable is on the params class.  Mostly just stylistic.

manifests/forward.pp
```diff
@@ -7,9 +7,7 @@
   $zone = $name,
 ) {
 
-  include unbound::params
-
-  $config_file = $unbound::params::config_file
+  $config_file = $unbound::config_file
```
xaque208: This is an out of scope variable unless we `include unbound`.

xaque208:
@t0mmyt Are you able to address the comments here?

igalic:
@t0mmyt ping

xaque208:
Closing PR due to lack of response.  If someone else wants to pick this work up, PRs welcome.

PR #84, opened 2015-04-23
restart unbound without starting and stoping the daemon

f0:
(default behaviour for..... e.g systemd), use the unbound-reload command also check the config before reload

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -64,6 +64,7 @@ class unbound (
   $val_clean_additional         = $unbound::params::val_clean_additional,
   $val_log_level                = $unbound::params::val_log_level,
   $val_permissive_mode          = $unbound::params::val_permissive_mode,
+  $validate_cmd                 = $unbound::params::validate_cmd,
   $verbosity                    = $unbound::params::verbosity,
   $custom_server_conf           = $unbound::params::custom_server_conf,
 ) inherits unbound::params {
@@ -87,6 +88,7 @@ class unbound (
     name      => $service_name,
     enable    => true,
     hasstatus => false,
+    restart   => 'unbound-control reload',
   }
 
   file { [
@@ -129,7 +131,8 @@ class unbound (
   }
 
   concat { $config_file:
-    notify  => Service[$service_name],
+    validate_cmd => $validate_cmd,
+    notify       => Service[$service_name],
   }
 
   concat::fragment { 'unbound-header':
--- a/metadata.json
+++ b/metadata.json
@@ -99,12 +99,12 @@
     },
     {
       "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": "3.6"
     }
   ],
   "description": "A module to deploy and manage the Unbound caching resolver",
   "dependencies": [
-    {"name":"puppetlabs/concat","version_requirement":">=1.1.0"},
+    {"name":"puppetlabs/concat","version_requirement":">=1.2.1"},
     {"name":"puppetlabs/stdlib","version_requirement":">=4.4.0"}
   ]
 }
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,7 +5,7 @@ fixtures:
       ref: '4.4.0'
     concat:
       repo: "https://github.com/puppetlabs/puppetlabs-concat.git"
-      ref: '1.1.0'
+      ref: '1.2.1'
     file_concat:
       repo: 'git://github.com/electrical/puppet-lib-file_concat.git'
       ref: '1.0.1'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -16,6 +16,7 @@ class unbound::params {
       $group            = 'unbound'
       $pidfile          = undef
       $fetch_client     = 'wget -O'
+      $validate_cmd      = undef
     }
     'redhat', 'centos', 'scientific': {
       $confdir          = '/etc/unbound'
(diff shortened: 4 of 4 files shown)


manifests/init.pp
```diff
@@ -87,6 +87,7 @@
     name      => $service_name,
     enable    => true,
     hasstatus => false,
+    restart   => 'unbound-checkconf && unbound-control reload',
```
igalic: and this works on *all* platforms?

igalic:
this conflates two things into reload (check and reload), when they are better done when (before) writing the configuration (with validate_cmd)

f0:
ok split the validation and reload makes sense.
what do you mean with all platforms? unbound-control comes with unbound, so i assume it works on all platforms

f0:
ok i implemented the split

manifests/init.pp
```diff
@@ -129,7 +130,8 @@
   }
 
   concat { $config_file:
-    notify  => Service[$service_name],
+    validate_cmd => 'unbound-checkconf %',
+    notify  	 => Service[$service_name],
```
igalic: please align these =>

metadata.json
```diff
@@ -99,12 +99,12 @@
     },
     {
       "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": "3.5"
```
igalic: puppet 3.5 was broken. it seems weird to explicitly depend on it

f0: hm ok don't know its broken, i depend on it because for validate_cmd minimal version is 3.5

igalic:
I'm curious what @xaque208 thinks about this

metadata.json
xaque208: 3.6 is the oldest place I'm using this module, so I'd be in favor of just going to 3.6 as the oldest Puppet version.  We'll just version this module accordingly.  The validate_cmd seems to rely only on the newer concat version unless I'm missing something.

xaque208: Ah, its the file resource in the concat module.  I'd say lets just go to 3.6.

xaque208: The docs state the following though:

"This command must have a fully qualified path, and should contain a percent (%) token where it would expect an input file. It must exit 0 if the syntax is correct, and non-zero otherwise."

That needs to be addressed before merge.

f0: ok i have adjustet the metadata.json to 3.6

2015-04-29 20:22 GMT+02:00 Zach Leslie <notifications@github.com>:

xaque208:
I'm generally +1 on this.  The validate_cmd needs to be worked out first.

xaque208:
Also, have you seen this one? https://github.com/xaque208/puppet-unbound/pull/70

f0:
yes, but its with an exec, i try to avoid execs....

2015-04-29 20:26 GMT+02:00 Zach Leslie <notifications@github.com>:

f0:
full qualified path is fixed

2015-04-29 20:28 GMT+02:00 Florian Koch <florian.koch1981@gmail.com>:

manifests/init.pp
```diff
@@ -129,7 +130,8 @@
   }
 
   concat { $config_file:
-    notify  => Service[$service_name],
+    validate_cmd => '/sbin/unbound-checkconf %',
```
xaque208: This needs to take into consideration other OSs.  On OpenBSD for example, this is '/usr/sbin/unbound-checkconf'.

xaque208:
I agree with avoiding execs, but having a $PATH to look in for the command should make some thing simpler.

f0:
yes thats true, but selectors are also an option, the default is /sbin and
each known other location needs to be defined

2015-04-29 20:34 GMT+02:00 Zach Leslie <notifications@github.com>:

xaque208:
Yep, and values that are missing for a given platform could just lack the validation functionality, perhaps.

f0:
hm yes but how to detect this? Wrapping the concat in an If block?

2015-04-29 20:53 GMT+02:00 Zach Leslie <notifications@github.com>:

xaque208:
I suspect that if validate_cmd => undef would solve it.  So as long as the variables looked up out of params had an undef if we don't know the path, that should handle it if we just passed that value to the concat resource.  I *think*.

f0:
ok do you man such a thing e.g

$unbound_validate_cmd = $osfamily ? {
    /OpenBSD/ => '/usr/sbin/unbound-checkconf %',
    /RedHat/  => '/sbin/unbound-checkconf %',
    default   => undef,
  }

  concat { $config_file:
    validate_cmd => $unbound_validate_cmd,
    notify       => Service[$service_name],
  }


2015-04-29 20:59 GMT+02:00 Zach Leslie <notifications@github.com>:

xaque208:
Yes, that kind of thing, though burying that in unbound::params.  That should make it easier, since the list of supported platforms is cased out in the params class.

f0:
ok i implement this tomorrow...

2015-04-29 21:14 GMT+02:00 Zach Leslie <notifications@github.com>:

f0:
hm i have no idea where the errors come from, on my  local test (rhel7/puppet .7.x) it works without errors

xaque208:
I don't think that failure is you.  I've put in https://github.com/xaque208/puppet-unbound/pull/85 to address changes in the concat module.

f0:
ok i wait for #85 merged and then run the tests again

xaque208:
Merged.  Rebase and lets see how it goes.

f0:
ok fixes also the version in fixture, now the build works

xaque208:
Cheers, thanks for working it out.

PR #87, opened 2015-05-08
add forward-first option for forward zones.

RyanFolsom:
Small change to allow for the forward-first option to be applied to forward zones.

### DIFF
--- a/manifests/forward.pp
+++ b/manifests/forward.pp
@@ -4,9 +4,13 @@
 #
 define unbound::forward (
   $address,
-  $zone = $name,
+  $zone          = $name,
   $forward_first = 'no'
 ) {
+
+  # Validate yes/no
+  $r_yesno = [ 'yes', 'no' ]
+  validate_re($forward_first, $r_yesno)
 
   include unbound::params
 
--- a/manifests/forward.pp
+++ b/manifests/forward.pp
@@ -5,6 +5,7 @@
 define unbound::forward (
   $address,
   $zone = $name,
+  $forward_first = 'no'
 ) {
 
   include unbound::params
--- a/templates/forward.erb
+++ b/templates/forward.erb
@@ -8,5 +8,5 @@ forward-zone:
   forward-addr: <%= @address %>
 <% end -%>
 <% if @forward_first != 'no' -%>
-  forward-first: <%= @forward_first -%>
+  forward-first: <%= @forward_first %>
 <% end -%>
--- a/templates/forward.erb
+++ b/templates/forward.erb
@@ -7,4 +7,6 @@ forward-zone:
 <% elsif @address != '' -%>
   forward-addr: <%= @address %>
 <% end -%>
+<% if @forward_first != 'no' -%>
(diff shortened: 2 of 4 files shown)


manifests/forward.pp
```diff
@@ -5,6 +5,7 @@
 define unbound::forward (
   $address,
   $zone = $name,
+  $forward_first = 'no'
```
xaque208: I'd like to see a validation around this for accepted values of 'yes' and 'no'.  Also please align `=`.

templates/forward.erb
```diff
@@ -5,5 +5,6 @@ forward-zone:
   forward-addr: <%= addr %>
 <% end -%>
 <% elsif @address != '' -%>
-  forward-addr: <%= @address -%>
+  forward-addr: <%= @address %>
 <% end -%>
+  forward-first: <%= @forward_first -%>
```
xaque208: Could we only include this line if the value of `@forward_first` is not the default?  This is really just for cleaner configs.

manifests/forward.pp
RyanFolsom: Can do. Is input validate being done elsewhere through the stdlib functions? Will I need to update the requirements section under README.md?

xaque208: Yes and yes.  The `validate_re()` function can help.

RyanFolsom: Want me to validate_array on $address while I'm in there? Looks like the template will only use it if it's an array.

xaque208: That also sounds like a good idea.

RyanFolsom: I was wrong about that - the template does work if it's not an array.

```puppet
<% if @address.is_a? Array -%>
<% @address.each do |addr| -%>
  forward-addr: <%= addr %>
<% end -%>
<% elsif @address != '' -%>
  forward-addr: <%= @address %>
<% end -%>
```
Sorry about that!

templates/forward.erb
```diff
@@ -5,5 +5,8 @@ forward-zone:
   forward-addr: <%= addr %>
 <% end -%>
 <% elsif @address != '' -%>
-  forward-addr: <%= @address -%>
+  forward-addr: <%= @address %>
+<% end -%>
+<% if @forward_first != 'no' -%>
+  forward-first: <%= @forward_first -%>
```
xaque208: The trailing - here in the erb may not be desired.

xaque208:
This looks reasonable.  If I'm wrong about the trailing -, this gets a +1 from me.

igalic:
you're right, @xaque208. that looks like my mistake. we don't need to propagate it.

RyanFolsom:
Following up - are there any other modifications you would like to see made to this PR?

xaque208:
Apologies for the delay, I'm just returning from holiday.  This looks good, thank you.

RyanFolsom:
No worries! Hope you enjoyed your holiday. Thank you for the merge!

PR #88, opened 2015-05-20
stub-zone could be specified with either ip or hostname

K-DEN:
This enables specifying stub zones with either or both of hostname(s) and ip address(s). We may want to create stub-zone settings in unbound.conf using ```stub-host:``` option in some case.

e.g. When using external nameserver services for some ip segment, it's likely that we would prefer specifying multiple name servers' hostnames to doing so with those ip addresses.

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -4,6 +4,7 @@
 #
 define unbound::stub (
   $address,
+  $hostname,
   $insecure = false
 ) {
 
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -4,13 +4,8 @@
 #
 define unbound::stub (
   $address,
-  $hostname,
   $insecure = false
 ) {
-
-  if ! $address and ! $hostname {
-    fail('unbound::stub: either address or hostname must be specified.')
-  }
 
   include unbound::params
 
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -10,6 +10,7 @@ define unbound::stub (
   if ! $address {
     fail('unbound::stub: address(es) must be specified.')
   }
+  unbound::stub::validate_addr{ $address: }
 
   include unbound::params
 
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -7,6 +7,10 @@ define unbound::stub (
   $insecure = false
 ) {
 
+  if ! $address {
+    fail('unbound::stub: address(es) must be specified.')
+  }
+
   include unbound::params
 
   $config_file = $unbound::params::config_file
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -8,6 +8,10 @@ define unbound::stub (
   $insecure = false
 ) {
 
+  if ! $address and ! $hostname {
+    fail('unbound::stub: either address or hostname must be specified.');
+  }
+
   include unbound::params
 
   $config_file = $unbound::params::config_file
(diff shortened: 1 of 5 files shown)


manifests/stub.pp
```diff
@@ -4,9 +4,14 @@
 #
 define unbound::stub (
   $address,
+  $hostname,
```
xaque208: Leaving no default values will force the user to set both of them, causing the failure below every time.

xaque208:
I like the idea, though the define needs some work.  Perhaps setting the default values to undef and requiring that at most one is set.

K-DEN:
Thank you for the review and comment. I intended that adding some flexibility might help in some case so that stubbing targets for one stub-zone can be specified with both of ip addresses and hostnames. Anyway the commit was incorrect in the sense that any stub-addr config can only be specified before any stub-hosts. So I will post another commit later. I would appreciate your reviewing.

templates/stub.erb
```diff
@@ -1,5 +1,11 @@
 stub-zone:
   name: "<%= @name %>"
+<% if @address -%>
 <% Array(@address).each do |addr| -%>
+<% if /\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\Z/ =~ addr -%>
```
xaque208: If this is meant to validate an IP address, I'd prefer that this get moved outside of the template.  Also, this test should take into consideration IPv6.

igalic: can't we just use function from stdlib?

xaque208: I expect yes.

K-DEN:
Thanks for the comments. I made another commit.
https://github.com/K-DEN/puppet-unbound/commit/987cf0362cb6eed2df30564b0165b82f2f2c0b0e
* separate defined resource type for address(es) validation added.
* for compatibility, unbound::stub args not changed.
* made it accomplish validation and to retain addresses order user provided, like in all the following cases,
    * address => ["{domain name 1}", "{ip address 1}", "{domain name 2}", ...]
    * e.g.
    ```
stub-addr: {domain name 1}
stub-host: {ip address 1}
stub-addr: {domain name 2}
:
```

* stub.erb template still contains is_domain_name. if you would come up with a nice way to remove it, I'd appreciate your help.

File: manifests/stub/validate_addr.pp [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+#
+define unbound::stub::validate_addr {
+  $addr_arr = split($name, '@')
+
+  # check if the address arg is a valid ip address
+  if is_ip_address($addr_arr[0]) {
+
+    # multiple '@' signs invalid
+    if size($addr_arr) > 2 {
+      fail('unbound::stub: invalid address. too many at signs in it.')
```
xaque208: Since we're testing for multiple '@', I like to see the use of '@' in the readme somewhere to give an indication how its used.

K-DEN:
Thank you for the comment. I made another commit. (one test case just for port validation)
https://github.com/K-DEN/puppet-unbound/commit/b1488bf039bbb0149c0e0773723132eda66bac50

manifests/stub/validate_addr.pp
K-DEN: Thank you. I added it so please see https://github.com/K-DEN/puppet-unbound/commit/b1488bf039bbb0149c0e0773723132eda66bac50 .

xaque208:
This is great, thank you.

PR #89, opened 2015-05-29
Some light fixes after recent merge

xaque208:
Two issues that needed addressing.

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -10,7 +10,9 @@ define unbound::stub (
   if ! $address {
     fail('unbound::stub: address(es) must be specified.')
   }
-  unbound::stub::validate_addr{ $address: }
+  unbound::stub::validate_addr{ "${address}_${name}":
+    address => $address,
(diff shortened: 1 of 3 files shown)


chrisboulton:
Doesn't this break the ability for `$address` to be an array?

It now creates one resource, and passes `$address` as an array into `unbound::stub::validate_addr` as a param. Previously the ability to have an array of addresses and validation for them depended upon Puppet creating a `unbound::stub::validate_addr` for each `$address` element.

You'll get an error like this:

    Error: Could not retrieve catalog from remote server: Error 400 on SERVER: private method `split' called for ["10.48.65.52@8600", "10.51.52.106@8600", "10.48.65.53@8600"]:Array at /opt/puppet_envs/consul/librarian-modules/unbound/manifests/stub/validate_addr.pp:9 on node xyz

xaque208:
It sure looks that way.  We need a test for this behavior.  Would you care to file an issue or send a PR?

PR #93, opened 2015-06-29
Bring back hide_version and hide_identity

### LINKED ISSUE #90: Missing support for hide_identity/hide_version
Support for hide_identity/hide_version got lost.  In commit df87e30, see file manifests/init.pp , line number 46 + 47. The $hide_identity and $hide_version variables were removed and never replaced with $unbound::params:: equivalents.

kmullin:
This patch should bring back parameters that were already in the template ([ref](https://github.com/xaque208/puppet-unbound/blob/a1704870ba23068ab19b8b1765641344d6d399c0/templates/unbound.conf.erb#L141-L150)), but would always evaluate to a false-ish value without these present.

The parameters were (I assume accidentally) removed in #54 

Closes #90


xaque208:
Cheers, thank you!  Yes accidentally :)

PR #94, opened 2015-06-29
stub: local-zones have multiple types; allow type to be overwritten

kmullin:
This patch introduces the ability to overwrite `transparent` as a stub-zone (local-zone)'s type.

I ran into this when trying to configure a different type for my `local-zone`.

From the [unbound.conf](https://www.unbound.net/documentation/unbound.conf.html) docs, a `local-zone` may have many types.

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -4,7 +4,8 @@
 #
 define unbound::stub (
   $address,
-  $insecure = false
+  $insecure = false,
+  $type     = 'transparent',
 ) {
 
   if ! $address {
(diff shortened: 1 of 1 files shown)


xaque208:
Thank you.

kmullin:
Actually looking at this, `unbound::stub` should probably make a call out to `unbound::local_zone` to utilize its already existing API for the proper `concat::fragment`.

Thoughts?

xaque208:
Ah, I don't use unbound::local_zone often so I forgot it was there but your probably right.  Is that a PR you're willing to send?

PR #95, opened 2015-07-03
Update control functionality

b4ldr:
The last update broke configurations that dont use the unbuond::remote class as control-enable was disabled resulting in the following error

```
Error: /Stage[main]/Unbound/Service[unbound]: Failed to call refresh: Could not restart Service[unbound]: Execution of 'unbound-control reload' returned 1: [1435927171] unbound-control[7677:0] warning: control-enable is 'no' in the config file.
[1435927171] unbound-control[7677:0] error: connect: Connection refused for 127.0.0.1
Error: /Stage[main]/Unbound/Service[unbound]: Could not restart Service[unbound]: Execution of 'unbound-control reload' returned 1: [1435927171] unbound-control[7677:0] warning: control-enable is 'no' in the config file.
[1435927171] unbound-control[7677:0] error: connect: Connection refused for 127.0.0.1
Wrapped exception:
Execution of 'unbound-control reload' returned 1: [1435927171] unbound-control[7677:0] warning: control-enable is 'no' in the config file.
[1435927171] unbound-control[7677:0] error: connect: Connection refused for 127.0.0.1
```

This attach allows for both configurations.  An error is still created when one switches from control-enable false -> true.  This happens because the catalog has control enable set to true, therefore the service reoload is set to use unbound-control however unbuond has not been restarted so it is not listening on the control port.  This results in the following error

```
Error: /Stage[main]/Unbound/Service[unbound]: Failed to call refresh: Could not restart Service[unbound]: Execution of 'unbound-control reload' returned 1: [1435929699] unbound-control[18018:0] error: connect: Connection refused for 127.0.0.1
Error: /Stage[main]/Unbound/Service[unbound]: Could not restart Service[unbound]: Execution of 'unbound-control reload' returned 1: [1435929699] unbound-control[18018:0] error: connect: Connection refused for 127.0.0.1
Wrapped exception:
Execution of 'unbound-control reload' returned 1: [1435929699] unbound-control[18018:0] error: connect: Connection refused for 127.0.0.1
```

Fixing this last error is somewhat chicken egg situation and any fix i could think of looked pretty messy


manifests/init.pp
```diff
@@ -90,7 +90,12 @@
     name      => $service_name,
     enable    => true,
     hasstatus => false,
-    restart   => 'unbound-control reload',
+  }
+  if $control_enable {
+    Service[$service_name] {
+      restart   => 'unbound-control reload',
+    }
+    include unbound::remote
```
xaque208: This looks the same as whats in https://github.com/xaque208/puppet-unbound/pull/97

b4ldr:
Yes it is, i tried to split them but i must have made a mistake when branching #97. Ill close this as you seem happy with #97

PR #96, opened 2015-07-03
Change scope of params

b4ldr:
One assumes unbound::remote would never be called without unbound.  Also the parameters  in remote should inherit any values which have been overridden by the user in unbound.  This request updates the params in remote so they inherit from unbound instead of just taking the static defaults from unbound::params


xaque208:
+1

PR #97, opened 2015-07-03
Freebsd 10

b4ldr:
As of freebsd 10 unbound is part of the base system.  This patch makes use of that instead of installing a second version of unbound


xaque208:
+1

xaque208:
Cheers, thank you.

PR #98, opened 2015-07-08
Multiple addressess/hosts for stub zones

rswarts:
With this change, we can once again have multiple addresses/hosts per stub zone.

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -11,8 +11,12 @@ define unbound::stub (
   if ! $address {
     fail('unbound::stub: address(es) must be specified.')
   }
-  unbound::stub::validate_addr{ "${address}_${name}":
-    address => $address,
+
+  # An ugly construction that alows us to validate $address in whatever way it
+  # is supplied.
+  $validatable_addresses = suffix(flatten([$address]), ",${name}")
+
+  unbound::stub::validate_addr{ $validatable_addresses:
   }
 
   include unbound::params
--- a/manifests/stub/validate_addr.pp
+++ b/manifests/stub/validate_addr.pp
@@ -2,11 +2,11 @@
 #
 # Validate address arg of unbound::stub
 #
-define unbound::stub::validate_addr(
-  $address
-) {
+define unbound::stub::validate_addr{
 
-  $addr_arr = split($address, '@')
+  $addr_name_arr = split($name, ',')
+
+  $addr_arr = split($addr_name_arr[0], '@')
 
   # check if the address arg is a valid ip address
   if is_ip_address($addr_arr[0]) {
@@ -22,7 +22,7 @@ define unbound::stub::validate_addr(
     }
 
   # address arg must be either ip address or hostname
-  } elsif ! is_domain_name($name) {
+  } elsif ! is_domain_name($addr_name_arr[0]) {
     fail('unbound::stub: invalid address.')
   }
 }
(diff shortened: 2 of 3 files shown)


sbadia:
Hi, :+1: otherwise we have an error… This a regression :-/
```puppet
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: undefined method `split' for ["10.42.1.221@10053", "10.42.1.211@10053"]:Array at /srv/puppet/production/modules/unbound/manifests/stub/validate_addr.pp
```

xaque208:
This change looks good to me, though I wish we had an rspec test to avoid further regression.

manifests/stub/validate_addr.pp
```diff
@@ -2,11 +2,11 @@
 #
 # Validate address arg of unbound::stub
 #
-define unbound::stub::validate_addr(
-  $address
-) {
+define unbound::stub::validate_addr{
 
-  $addr_arr = split($address, '@')
+  $addr_name_arr = split($name, ',')
```
xaque208: In what circumstance would a namevar received here contain a comma?

manifests/stub.pp
```diff
@@ -11,8 +11,12 @@
   if ! $address {
     fail('unbound::stub: address(es) must be specified.')
   }
-  unbound::stub::validate_addr{ "${address}_${name}":
-    address => $address,
+
+  # An ugly construction that alows us to validate $address in whatever way it
+  # is supplied.
+  $validatable_addresses = suffix(flatten([$address]), ",${name}")
```
xaque208: Why append the comma-name to each element?

xaque208:
I've sent https://github.com/rswarts/puppet-unbound/pull/1 which if merged, will handle the base spec test I'm after.  I think I have it correct, this code doesn't seem to work for '10.0.0.1@10053' style addresses, as we have in the readme.  I'd still like clarification around why we append the ,$name to each element.  Is that just to avoid name space collisions?

manifests/stub/validate_addr.pp
rswarts: The comma is introduced in the namevar in manifests/stub.pp:

```puppet
$validatable_addresses = suffix(flatten([$address]), ",${name}")
```

manifests/stub.pp
rswarts: To make sure all the unbound::stub::validate_addr resources have a unique name.

If we have multiple stub zones with the same addresses, as such:
```puppet
unbound::stub {
  'example.com':
    address => ['10.19.0.3', '10.19.1.3' ];
  'example.net':
    address => ['10.19.0.3', '10.19.1.3' ];
}
```

this will result in

```puppet
$validatable_addresses = [
  '10.19.0.3,example.com',
  '10.19.1.3,example.com',
  '10.19.0.3,example.net',
  '10.19.1.3,example.net'
]
```

rswarts:
Hi @xaque208, your test works fine, just had a small typo in it.

The reason for the ugly ,$name appending is indeed preventing two stub zones with the same addresses causing problems. Without involving the future parser and making everyone not using it unhappy, I could not find a nicer way to make this work for both single and multiple addresses for multiple stub zones.

xaque208:
Yep, I figured it out after I continued staring at it. :)  It'd be nice to move the validation logic into a function or some such to clean up the code some.  That sounds like another day's work though.

This all looks good.  Thank you for the efforts, and sorry for the breakage.

PR #101, opened 2015-07-20
Fix dependencies when using unbound::remote

dvorak:
When first applying the manifest it's possible for the unbound::remote
class to be evaluated before the packages are installed or for the
service to be started before the remote management is configured.  This
addresses both issues.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -96,7 +96,7 @@ class unbound (
       restart   => 'unbound-control reload',
       require   => Class['unbound::remote'],
     }
-    Package[$package_name] -> Class['unbound::remote']
+    Package<| title == $package_name |> -> Class['unbound::remote']
     include unbound::remote
   }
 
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -94,7 +94,9 @@
   if $control_enable {
     Service[$service_name] {
       restart   => 'unbound-control reload',
+      require   => Class['unbound::remote'],
     }
+    Package[$package_name] -> Class['unbound::remote']
```
xaque208: Not all platforms have a package to install, so we'll need to provide some logic to work around that.

dvorak:
Updated to make package dependency optional

manifests/init.pp
```diff
@@ -94,7 +94,9 @@
   if $control_enable {
     Service[$service_name] {
       restart   => 'unbound-control reload',
+      require   => Class['unbound::remote'],
     }
+    Package<| title == $package_name |> -> Class['unbound::remote']
```
xaque208: Will this require Package[$package_name] to be in the catalog?

dvorak: Nope, it there is no match then it will not create a dependency.  It will silently "fail".

xaque208:
Nice, I just tested out.  Catalog compiles.  Thanks for the contribution.

Issue #102, opened 2015-07-21
Stub addresses are generated as hostnames in config


F00baer:
Hello, 
I had the problem that IPv4 addresses were generated as "stub-host" entries in the config file instead of "stub-addr". 
Maybe it is better to check for valid IP instead of domain_name in the stub template?

<% if scope.function_is_ip_address([addr]) -%>
  stub-addr: <%= addr %>
<% else -%>
  stub-host: <%= addr %>
<% end -%>

xaque208:
Can I see the manifest you used to get into the situation you mentioned?

F00baer:
On 2015-07-23 17:35, Zach Leslie wrote:
> Can I see the manifest you used to get into the situation you mentioned?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/xaque208/puppet-unbound/issues/102#issuecomment-124142224>.
> 


Sure,

I manage my stubs as hiera data, like this:

recursor::stubs:
  domain1.com:
    address:
      - "121.120.1.64"
      - "121.120.6.5"
      - "121.120.6.6"
      - "2001:123:4:6::88:5"
      - "2001:123:4:6::88:6"
    insecure: true
  domain2.com:
    address:
      - "121.121.1.64"
      - "121.130.6.11"
      - "121.130.6.12"
      - "2001:132:4:6::88:11"
      - "2001:123:4:6::88:12"
    insecure: true



The class is defined in a profile:

  class { "unbound":
    interface => $ips,
    access    => $access,
    do_ip6    => $do_ip6,
    do_ip4    => "yes",
    hide_identity => false,
    hide_version => false,
    val_clean_additional => false,
    control_enable => true,
    num_threads => $processorcount,
    msg_cache_slabs => $processorcount*2,
    rrset_cache_slabs => $processorcount*2,
    infra_cache_slabs => $processorcount*2,
    key_cache_slabs => $processorcount*2,
    outgoing_interface => $ipaddress,
    prefetch => "yes",
    so_rcvbuf => "4m",
    num_queries_per_thread => "4096",
    outgoing_range => 8192,
    rrset_cache_size => "${rcache}m",
    msg_cache_size => "${mcache}m",
    edns_buffer_size => "4096",
    outgoing_port_avoid => "0-1024",
    outgoing_port_permit => "1025-65535",
    val_permissive_mode => $permissive_mode,
    module_config => $module_config,
    interface_automatic => $interface_automatic,
    chroot => $chroot,
[...]

PR #104, opened 2015-07-28
Begin replacing unbound::stub address validation

xaque208:
This work moves the validation out of Puppet and into Ruby.  This should
allow for a more easily tested and flexible validation using the
'ipaddr' core library in Ruby as of at least 1.9.3.


xaque208:
I've tested and had confirmation in #102.

PR #106, opened 2015-10-13
Add feature skip_roothints_download

### LINKED ISSUE #103: Option not to use/download root.hints
Having a purely caching forwarder which just forwards request to some other recursive nameserver there is no need for the root.hints file and in our case the servers can't even reach the Internet so it hangs on downloading the root.hints (until touching it manually). Please create an option not to use/download root.hints.

mrdima:
An option to skip the download of root.hins. There are many ways, I wanted to change as little as possible. Closes #103

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,6 +69,7 @@ class unbound (
   $validate_cmd                 = $unbound::params::validate_cmd,
   $verbosity                    = $unbound::params::verbosity,
   $custom_server_conf           = $unbound::params::custom_server_conf,
+  $skip_roothints_download      = $unbound::params::skip_roothints_download,
 ) inherits unbound::params {
 
   if $package_name {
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -100,6 +102,12 @@
     include unbound::remote
   }
 
+  if $skip_roothints_download {
+    File[$hints_file] -> Exec['download-roothints']
+  } else {
+    Exec['download-roothints'] -> File[$hints_file]
+  }
```
igalic: i like this hack ;)

igalic:
i was gonna say: can you add a spec tests, but we don't have any spec tests for this class now (other than  whether it exists), so it wouldn't be fair on you…

you should however document the feature in the readme!

mrdima:
Yes, I thought about the readme after submitting, I'll submit that. Yeah, rspec and even general parameter validation is lacking in this module

mrdima:
I tried to keep the documentation in the style of the readme, it's a bit different from regular modules which usually document all the parameters, hope this is ok.

igalic:
super! thank you @mrdima!

PR #107, opened 2015-12-06
Use the relative path for loading puppet_x

xaque208:
This is to accommodate Puppetserver that doesn't currently assist in
dependency resolution.


xaque208:
Tested functional.

PR #109, opened 2016-01-12
Allow to pass action with $access

sileht:
Hi,

This is about allowing to replace allow access with any other values like allow_snoop, denied, reject...

Cheers,


templates/unbound.conf.erb
```diff
@@ -81,11 +81,19 @@ server:
 <% end -%>
 <% if @access.is_a? Array -%>
 <% @access.each do |acc| -%>
+<% if acc.chomp().split(' ').length != 1 -%>
+  access-control: <%= acc %>
+<% else -%>
```
igalic: i'm very confused by this code

sileht: That means we don't add 'allow' if 'acc' have more than 1 words.

This allows to do thing like that :

access => ['127.0.0.1/32 allow_snoop', '124.69.127.0/22 allow', '124.69.127.0/22 deny']

xaque208: Could check for whitespace instead of splitting, but that's minor.  I'd think this is the kind of thing that would be suited for a test, as well as a readme update.  Readme for sure, so users who want to use it don't have to read code.

sileht: I have done that.

igalic:
@sileht what you actually want to update are the tests in [spec/](https://github.com/xaque208/puppet-unbound/blob/master/spec/classes/init_spec.rb)

@xaque208 we should rename `tests` to `examples`

sileht:
I have added specs tests and rename the examples directory

igalic:
thank you so much @sileht!

PR #111, opened 2016-01-27
Add hiera support for define resources

Rocco83:
Hi,
I have added, as it is in apt module, the logic in order to handle definition through hiera.

Regards,
Daniele

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -56,6 +56,16 @@ At minimum you should setup the interfaces to listen on and allow access to a fe
     }
 ```
 
+Or, using hiera
+```yaml
+unbound::interface:
+  - '::0'
+  - '0.0.0.0'
+unbound::access:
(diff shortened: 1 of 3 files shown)


xaque208:
This looks good, thanks for the PR.  How do you feel about a test or three?  This looks like the kind of thing that would be relatively simple to spec test.  The data doesn't need to come from hiera, but could just be passed in at the `unbound` class.

Rocco83:
Hi, i didn't had time yet. I will do it asap.
Cheers

Rocco83:
Hi,
I have created one test for each parameter.
I have choosed this path in order to let the failure happens on one single test.
If you have any suggestion, feel free to do it!

HTH,
Daniele

xaque208:
This is good work, thank you for the efforts.

PR #112, opened 2016-01-27
using https to download named.root

mmckinst:


xaque208:
This seems reasonable, though I think it could use a test on on a brand new system.  Specifically, I don't know if unbound packages on the majority of systems bring in the ca-certificates package (or whatever its called) so this might break the fetch.  I'm also unclear if this increases security, considering internic allows fetching over ftp `ftp://ftp.internic.net/domain/named.cache`.

mmckinst:
if someone wants to use download the hints file over http or ftp they still can, I just think it would be a much better to default to https.

the hints file is downloaded using wget, curl, or fetch depending on the distro. the ca bundle should be a requirement wget, curl, or fetch. I can't imagine a distro installing curl or wget (both do SSL/TLS) but not including the ca bundle.

xaque208:
This seems fine.  I looked at the dependencies for curl on FreeBSD and wget on Debian.  Both will drag in the ca certificate bundle by default.  Thanks for the PR.

PR #113, opened 2016-01-29
Added missing option in the configuration: so_sndbuf

marknl:
I am using this parameter and was missing it from the parameters list. I know there's a custom_server_conf option, but since you did declare so_rvcbuf, you should also define it's counterpart for consistency ;-)


xaque208:
Thank you.  Would you mind squashing those commits down please?

marknl:
Voila.

xaque208:
Thank you for the contribution.

PR #114, opened 2016-02-21
 Convert config_file to a definition parameter

sbadia:
Hi here!

For the moment we can't define a different `config_file` param in `::unbound` and
configure another `::unbound::forward` or `stub` using defined type in the same
catalog. Because the dependency is not satisfied, and lead to an error.

```puppet
Error: Failed to apply catalog: Could not find dependent Exec[concat_/etc/unbound/unbound.conf] for File[/var/lib/puppet/concat/_etc_unbound_unbound.conf/fragments/15_unbound-stub-42.10.in-addr.arpa.] at /srv/puppet/production/modules/concat/manifests/fragment.pp:132
```

my `manifest` is pretty simple:

```puppet
  class {
    '::unbound':
      config_file => "${::unbound::params::confdir}/unbound.conf.d/toto.conf",
      access      => [
        '::1',
        '127.0.0.1/8',
        'x.x.x.x/xx',
        'xxxx:xxx::/xx',
      ],
    }

    unbound::stub { '42.10.in-addr.arpa.':
      address  => ['x.x.x.x@53','xxxx:xxx::x@53'],
      insecure => true,
    }
```


xaque208:
Hi!  This is good work.  Thank you for the contribution.

sbadia:
@xaque208 thanks!

PR #116, opened 2016-03-31
Puppet4

mmckinst:


mmckinst:
Its failing because of the problem [described here](https://tickets.puppetlabs.com/browse/PUP-5683?focusedCommentId=253446&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-253446) . `$::operatingsystemrelease` isn't described anywhere explicitly in the tests. The same is done for OpenBSD in params.pp

Considering FreeBSD <= 10.0 is EOL, the conditional could be removed but that will effectively drop support for FreeBSD because metadata.json only has versions <= 10.0 . Has this been tested on newer versions of FreeBSD?

xaque208:
Yes, I have this module running on FreeBSD 10.2, 10.3-RC2 and OpenBSD 5.8.  I think we can deprecate old.  I'll keep that in mind when I bump the version.

xaque208:
To clarify, are you planning to drop the support for older versions of BSD in this PR?

mmckinst:
I'm dropping old versions and adding the new versions you listed.

mmckinst:
I think there's a bug in the current version. It expects a line to look like this `local-data: \"a.example-record.com 14400 IN A 10.0.0.110.0.0.2` but there should be a space between those IPs. I updated it in 00f1a1a . Let me know if you want these commits squashed.

xaque208:
Hmm, I see the bug.  Thanks for the fix.

Though, is that a valid configuration?  My question is more about, is it legal configuration to have multiple IP addresses on a "resource record string"?  That looks strange to me, but I could be wrong.

mmckinst:
I'm not sure either actually, I've never used that feature of it. Do you think it would be better to just split it across multiple lines?

xaque208:
I did a bit more reading on the `unbound.conf(5)` page, as well as a local install and am more convinced that this (multiple addresses on a single local-data line) is not the correct behavior.  I also recognize that this isn't your bug.  Though, it will take me a little while to fix this if nobody else gets to it.

In the world of future parser, or puppet4, we could just loop over the array in the `unbound::record` class, but that would break backwards compatibility.  Personally I'm completely migrated to puppet4, and we could cut a backwards incompatible release but that doesn't have to be here.  Perhaps in the template makes the most sense for now.

xaque208:
So, I thought about this again, and I think the short answer here is to merge https://github.com/xaque208/puppet-unbound/pull/117 and close this one.

What do you think @mmckinst ?

mmckinst:
I think thats fine. FWIW, I'd be fine with dropping puppet 3 but I know lots of other people are still using it now so it might not be practical.

xaque208:
Lets start here and we can take an approach for dropping puppet3 as its own item.

PR #121, opened 2016-04-14
fix extended_statistics setting

mmckinst:
Right now `$extended_statistics` is set to the string `no`, this is put
in the the config file using `extended-statistics: <%=
@extended_statistics %>` .

If you try to set `$extended_statistics = yes` or `extended_statistics =
'yes'` in hiera, YAML(?) casts that as a boolean and the config file
gets generated as `extended-statistics: true` which causes unbound to
fail because it expects a `yes` or `no`.


xaque208:
Great, thank you for the contribution.

PR #122, opened 2016-05-07
booleans: allow to actually toggle the booleans

igalic:
some of these options not only default 'yes' in our config but in the
daemon's internals, thus, not putting it in the config means we cannot
disable them.


xaque208:
Thank you @igalic

PR #124, opened 2016-06-02
Add ability to set cache settings as parameters

jaxxstorm:
This MR adds the ability to set two very widely modified config settings, `cache_max_ttl` and `cache_max_negative_tt`l.

While these can be set use `custom_server_conf` but I think these options are commonly enough that it's justifiable to add these as explicit options.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -72,6 +72,7 @@ class unbound (
   $val_permissive_mode          = $unbound::params::val_permissive_mode,
   $validate_cmd                 = $unbound::params::validate_cmd,
   $verbosity                    = $unbound::params::verbosity,
+  $cache_max_ttl                = $unbound::params::cache_max_ttl,
   $custom_server_conf           = $unbound::params::custom_server_conf,
   $skip_roothints_download      = $unbound::params::skip_roothints_download,
 ) inherits unbound::params {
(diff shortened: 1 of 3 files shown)


xaque208:
This looks good, thank you for the PR.  Would you mind ordering the params to be alphabetical?  I think there is one below your change that is out of place, but everything else looks ordered.

jaxxstorm:
Can do, but I realised that `cache_max_negative` is only available from later versions of unbound.

Perhaps it's best to leave it as a custom server conf option with that in mind. Considering.

xaque208:
@jaxxstorm Do you have a link to release notes about the inclusion of `cache_max_negative` or a note about when it was introduced?  If the platforms we support in this module contain those versions, I'd think its reasonable to include here as well.

jaxxstorm:
https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=675

xaque208:
Given that the default here is to use undef, thus not placing an entry in the resulting on-disk configuration file, it seems fine to merge here with the idea that folks should know what version of unbound they are using, and which options are supported.  Trying to track down the matrix of unbound version + supported options + platform packing is just too much.

I'm +1 to merge here, but I would take arguments from others if they feel different.

igalic:
*agreed*

thanks @jaxxstorm for the patch

PR #130, opened 2016-06-11
Set hints permissions after download

dvorak:
The package will usually contain a hints file, but the module will
download a new one.  We need to make sure the hints file is managed
after it's downloaded so that the runs are idempotent.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -149,8 +149,9 @@ class unbound (
   }
 
   file { $hints_file:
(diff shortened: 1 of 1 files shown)


xaque208:
Thank you for the PR.  Did you intend to remove the `mode` param from the file resource here?

dvorak:
The mode change was accidental.  Should be fixed now.

igalic:
thanks @dvorak!

PR #131, opened 2016-06-23
Remove dependency cycle when skip_roothints_download is true

drt24:
The requires for download-roothints and the hints file are set up earlier in
the file conditional on $skip_roothints_download specifying it again here
causes a dependency cycle.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -149,8 +149,8 @@ class unbound (
   }
 
   file { $hints_file:
-    ensure  => file,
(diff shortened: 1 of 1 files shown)


xaque208:
This change looks good since the resource relationships are captured in the conditional above.  The tests are failing though due to an extra whitespace now that the parameter is removed.  Would you mind correcting that please?

xaque208:
Excellent, thank you.

PR #132, opened 2016-07-20
Begin migration away from params

xaque208:


zleswomp:
Passing tests are cool.  Its a start.

xaque208:
The undef value defaults here need to be addressed before this is ready for commit.

igalic:
***nice***

n.b.: in @voxpupuli we have [tea](https://github.com/voxpupuli/puppet-tea) some composite types such as Port  and Unprivileged Port and UnixPath and so on

zleswomp:
Oh sweet!  Thanks for the pointer @igalic.  I'll drag some of that in for path validation etc.  It will probably still take a few more passes over this to get it in shape for merge.

igalic:
@zleswomp please note: prs to extend tea are very warmly welcomed!

xaque208:
Dear me, come back to this.

xaque208:
I think some of the data types validation will have to wait here.  What I'd like to do is replace the lib code here for IP validation with validation from native Pupept 4 types.  The IP address can come from thrnioi/ip, and the port can come from voxpupuli/tea, but combining those two looks like more work than I'd like to put in right now.  I've gotten this code running stable in my environment, so I think some sanity checking here would be good.  I'll come back to pass over this PR later.

If folks from the community would like to review this code before merge, I think this would be a good time for that.  I'll plan to squash up my commits before merge.

xaque208:
Updating the supported OS means that the fact sets for FreeBSD are actually being loaded now, which has uncovered a few bugs that need addressing.  I'll work through those next.

xaque208:
I think this is about as ready as its going to get for now.  Review would be appreciated.  I'll hold off merging a day or so.

metadata.json
```diff
@@ -73,14 +73,14 @@
     {
       "operatingsystem": "FreeBSD",
       "operatingsystemrelease": [
-        "10.1",
-        "10.2"
+        "10"
```
igalic: 11 is out… does it still work there?

spec/classes/init_spec.rb
```diff
@@ -1,129 +1,145 @@
...
-    ].each do |operatingsystem|
-      describe "puppet class without any parameters on #{operatingsystem}" do
-        let(:facts) {
-          {
-            :operatingsystem => operatingsystem,
-            :concat_basedir => '/dne'
-          }
-        }
-        it { should contain_class('unbound') }
+  on_supported_os.each do |os, facts|
```
igalic: :+1:

metadata.json
xaque208: It does, thanks.  Added.

PR #135, opened 2016-11-11
Parameter to toggle DNSSEC

ct16k:
The patch adds a Boolean $disable_dnssec (defaults to false), which prevents the trust anchors from being added to the config file, thus disabling DNSSEC. This is needed in a setup where validation is unfortunately still unavailable and unbound returns SERVFAIL with it enabled. According to [docs](https://www.unbound.net/documentation/howto_turnoff_dnssec.html), permissive mode doesn't actually disable validation, but ignores the result, so could be viewed as a performance hit in shops not using DNSSEC. 

I've created a 1.x branch to add the flag to the old syntax version as well, but I don't know if that is still supported.

### DIFF
--- a/templates/unbound.conf.erb
+++ b/templates/unbound.conf.erb
@@ -2,8 +2,10 @@
 #
 server:
   verbosity: <%= @verbosity %>
+<% if !@disable_dnssec -%>
   trusted-keys-file: <%= @trusted_keys_file %>
   auto-trust-anchor-file: <%= @auto_trust_anchor_file %>
+<% end -%>
   do-not-query-localhost: no
   use-syslog: yes
 <% if @extended_statistics -%>
@@ -160,7 +162,7 @@ server:
 <% else -%>
   val-clean-additional: no
 <% end -%>
-<% if @val_permissive_mode -%>
+<% if @val_permissive_mode && !@disable_dnssec -%>
   # NOTE: TURNING THIS ON DISABLES ALL DNSSEC SECURITY
   val-permissive-mode: yes
 <% end -%>
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -12,6 +12,7 @@ unbound::validate_cmd: '/usr/sbin/unbound-checkconf %'
 unbound::access:
   - '::1'
   - '127.0.0.1/8'
+unbound::disable_dnssec: false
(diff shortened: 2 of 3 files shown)


templates/unbound.conf.erb
```diff
@@ -160,7 +162,7 @@ server:
 <% else -%>
   val-clean-additional: no
 <% end -%>
-<% if @val_permissive_mode -%>
+<% if @val_permissive_mode && !@disable_dnssec -%>
```
xaque208: If both `@val_permissive_mode` and `@disable_dnssec` are true here, than the line for `permissive` doesn't get included in the config.  Is that the desired outcome here?

xaque208:
Thank you for the PR.  The motivation here sounds good to me, but I'm less excited about the specific template change.  Just looking it over, it seems to me that, instead of adding a new parameter for disabling DNS sec, if we wrapped conditions in the template around `trusted-keys-file` and `auto-trust-anchor-file`, we would have ability to set those to `undef` or `false` or `''` that would allow those items to be skipped in the configuration.  Then adding a section to the readme would be all that is required to allow users to know how to disable DNSSEC.

I've commented on the line that I find a bit confusing.  If we do in fact want a single parameter to disable DNSSEC as this PR stands currently, I think perhaps a test is in order to ensure that the logic for the variables in play here result in a the desired config.  Would you be willing to add those tests for this change?

templates/unbound.conf.erb
b4ldr: i dont think it makes much difference the removal of the trust anchors file disables dnssec

https://unbound.net/documentation/howto_turnoff_dnssec.html

b4ldr:
Another option would be to support `module-config:` . something like the following

```puppet
class unbound (
  Array[Enum['iterator', 'validator']] $module_config = ['validator ', 'iterator']
) {
}
```

unound.conf.erb
```erb
server:
  <%# we need the sort to ensure validator is first %>
  module-config: <%= @module_config.sort.reverse.join %>
```

alternatively we could have an enable_validator parameter, im not sure if it makes sense to have a disable_iterator paramter

https://unbound.net/documentation/howto_turnoff_dnssec.html

b4ldr:
https://github.com/xaque208/puppet-unbound/pull/176 adds module_config which allows you to disable the validator

PR #141, opened 2017-06-21
Updated to include python support

mlosapio:
Adding python support for 1.3.6 line

### DIFF
--- a/templates/unbound.conf.erb
+++ b/templates/unbound.conf.erb
@@ -182,3 +182,7 @@ server:
 <% @custom_server_conf.each do |c| -%>
   <%= c %>
 <% end -%>
+<% if @python -%>
+python:
+  python-script: "<%= @python %>"
+<% end -%>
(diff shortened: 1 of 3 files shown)


templates/unbound.conf.erb
```diff
@@ -182,3 +182,7 @@ server:
 <% @custom_server_conf.each do |c| -%>
   <%= c %>
 <% end -%>
+<% if @python -%>
+python:
```
xaque208: Is this the correct syntax?  Looking in the manpager makes it seem like `python-script` is the top level key here.

mlosapio: Yeah, the docs are a little weak but this is correct: 
https://github.com/jedisct1/unbound/blob/master/doc/example.conf.in#L737

xaque208:
Would you mind rebasing here please?

xaque208:
Hi @mlosapio .  If this is something you'd still like to get added to the module, would you mind rebasing please?  There have been some significant changes to that template since this has been open.

b4ldr:
https://github.com/xaque208/puppet-unbound/pull/176 adds support for the python module

PR #142, opened 2017-09-18
Add common options

ahotton:
add different common options


manifests/init.pp
```diff
@@ -13,6 +13,8 @@
   String $conf_d,
   String $confdir,
   String $config_file,
+  Array $control_interface,
```
xaque208: This one doesn't look to get used, though, I see `$control_port` in the template.  A mixup perhaps?

xaque208:
Thank you for the PR.  What is the result of the do-not-query-address defaulting to localhost?

data/common.yaml
```diff
@@ -16,12 +16,19 @@ unbound::record: {}
 unbound::access:
   - '::1'
   - '127.0.0.1/8'
+unbound::do_not_query_address:
```
xaque208: Are localhost addresses the correct default here?  We don't currently have anything set, so this might be a surprise for folks who run a authoritative server locally on another port perhaps.

PR #143, opened 2017-10-05
Begin testing for confdir permission

xaque208:


xaque208:
The tests here are not being run due to facterdb support.

PR #144, opened 2017-10-10
Add more options

ahotton:


xaque208:
Thank you for the PR.  The errors in the tests here look to be related to the default values without being specified as `undef` in `common.yaml`.

Issue #146, opened 2017-11-09
Add option to allow unbound to log to file instead of syslog


enemarke:

enemarke:
#147

xaque208:
#147 has been merged.  Thank you.

PR #147, opened 2017-11-09
Added option to allow unbound to log to file instead of syslog

enemarke:
#146


xaque208:
Thank you for the contribution.

PR #149, opened 2017-12-14
Add more options

ahotton:
Adding the following options :
- Domain insecure
- Cache min ttl
- Control interface
- Do not query address


xaque208:
Hi @ahotton, thanks for the PR.  Would you mind squashing this down?  A `git rebase -i master` should get you close.

PR #150, opened 2017-12-20
add new options

ahotton:
Add following options: 

do-not-query-address
cache-min-ttl
domain-insecure


data/common.yaml
```diff
@@ -76,8 +79,10 @@ unbound::val_clean_additional: true
 unbound::val_log_level: 1
 unbound::val_permissive_mode: false
 unbound::verbosity: 1
+unbound::cache_min_ttl: ~
 unbound::cache_max_ttl: ~
 unbound::cache_max_negative_ttl: ~
 unbound::custom_server_conf: []
 unbound::skip_roothints_download: false
+unbound::domain_insecure: undef
```
xaque208: This `undef` should be `~`.

manifests/init.pp
```diff
@@ -13,6 +13,8 @@
   String $conf_d,
   String $confdir,
   String $config_file,
+  String $domain_insecure,
```
xaque208: In order to allow the `~` option above, this will need to change to `Optional[String] $domain_insecure`.

ahotton:
@xaque208 it's Ok

templates/unbound.conf.erb
```diff
@@ -10,6 +10,13 @@ server:
   use-syslog: no
 <% else -%>
   use-syslog: yes
+<% if @do_not_query_address.is_a? Array -%>
```
xaque208: We've set the data type of `do_not_query_address` to be an array, so we probably don't need to check the data type.

data/common.yaml
```diff
@@ -16,6 +16,9 @@ unbound::record: {}
 unbound::access:
   - '::1'
   - '127.0.0.1/8'
+unbound::do_not_query_address:
```
xaque208: What does it mean to have this default in the config?  Since you're chaning the default behavior, I'll need an appropriate version bump here, but more than that, I'd like to understand what it means to include both the local addresses to the `do_not_query_address`.

ahotton:
@xaque208 
Hello,
I changed the behavior of do-no-query-address in optional to limit the impact on default configuration.
I use this option to restrict the communication surface to maximum.

templates/unbound.conf.erb
```diff
@@ -11,6 +11,9 @@ server:
 <% else -%>
   use-syslog: yes
 <% end -%>
+<% if @do_not_query_address -%>
+  do-not-query-address: <%= @do_not_query_address%>
```
xaque208: I think we still need the array loop around this in order to allow multiple values in the array, when its set.

ahotton:
@xaque208 

I changed it.

xaque208:
This looks good @ahotton.  Would you mind squashing down the commits please?

manifests/init.pp
```diff
@@ -13,6 +13,8 @@
   String $conf_d,
   String $confdir,
   String $config_file,
+  Optional[String] $domain_insecure,
```
b4ldr: this should be an array so one can provide multiple domains

templates/unbound.conf.erb
```diff
@@ -11,6 +11,13 @@ server:
 <% else -%>
   use-syslog: yes
 <% end -%>
+<% if @do_not_query_address -%>
+<%- @do_not_query_address.each do |int| -%>
+  do-not-query-address: <%= int %>
+<%- end -%>
+<% elsif @do_not_query_address != '' -%>
```
b4ldr: this will never be a string unless you change the type definition to `Optional[Variant[Array,String]] $do_not_query_address` 
personally i would keep things as they are so it only supports arrays and remove this clause

xaque208: Thanks for pointing that out.

manifests/init.pp
xaque208: That also sounds like a good change.

PR #151, opened 2018-01-23
Define local_zone via hiera

b4ldr:
Currently you can define stubs, forwards and access types via hiera however the local_zone type was missed.  this pull addresses that.  I have also updated the README, spec tests and added a custome type to validate the local-zone type

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,32 @@ This means that your server will use the Google DNS servers for any
 zones that it doesn't know how to reach and cache the result.
 
 
+### Local Zones
+
+Configure a local zone. The type determines the answer  to  give
+if  there  is  no  match  from  local-data.  The types are deny,
+refuse, static, transparent, redirect, nodefault,  typetranspar-
+ent,  inform,  inform\_deny,  always\_transparent,  always\_refuse,
+always\_nxdomain.  See local-zone in the [unbound documentation](https://unbound.net/documentation/unbound.conf.html) 
+for more information.  You can configure a local-zone with something like the 
+following.
+
+```puppet
+puppet::local_zone {'10.0.10.in-addr.arpa.':
+  type => 'nodefault'
+}
+```
+
+Or, using hiera
+```yaml
+unbound::local_zone:
+  10.0.10.in-addr.arpa.:
+    type: nodefault
+  11.0.10.in-addr.arpa.:
+    type: nodefault
+```
+
+
 ### Fine grain access-control
 
 ```puppet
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -148,6 +148,32 @@ This means that your server will use the Google DNS servers for any
...
 
+### Local Zones
+
+Configure a local zone. The type determines the answer  to  give
+if  there  is  no  match  from  local-data.  The types are deny,
+refuse, static, transparent, redirect, nodefault,  typetranspar-
+ent,  inform,  inform\_deny,  always\_transparent,  always\_refuse,
+always\_nxdomain.  See local-zone in the [unbound documentation](https://unbound.net/documentation/unbound.conf.html) 
+for more information.  you can configure a local-zone with something like the 
+following
```
xaque208: This one needs a trailing period and a capital.

README.md
```diff
@@ -148,6 +148,32 @@ This means that your server will use the Google DNS servers for any
 zones that it doesn't know how to reach and cache the result.
 
 
+### Local Zones
+
+Configure a local zone. The type determines the answer  to  give
+if  there  is  no  match  from  local-data.  The types are deny,
+refuse, static, transparent, redirect, nodefault,  typetranspar-
+ent,  inform,  inform\_deny,  always\_transparent,  always\_refuse,
+always\_nxdomain.  See local-zone in the [unbound documentation](https://unbound.net/documentation/unbound.conf.html)
```
xaque208: Thanks for the link to the docs.

manifests/local_zone.pp
```diff
@@ -18,9 +18,9 @@
 #   (optional) name of configuration file
 #
 define unbound::local_zone (
-  $type,
-  $zone = $name,
-  $config_file = $unbound::config_file,
+  Unbound::Local_zone_type $type,
```
xaque208: Nice one.

b4ldr:
ok updated.  Turns out what i actually needed was unbound::stub.  which worked a bit differently then expected and made me ponder:
  1. is there any reason you would want to use unbound::local_zone instead of unbound::local_zone?
  2. why doesn't `unbound::stub` use `unbound::local_zone` instead of https://github.com/xaque208/puppet-unbound/blob/master/manifests/stub.pp#L58-L62
  3. should there be an `unbound::domain_insecure` and if so should one call that from `unbound::stub`
  4. perhaps we could create an `unbound::stub_zone` and just have `unbound::stub` as a wrapper to the others?

PR #152, opened 2018-01-23
refactor local-zone

b4ldr:
after reflecting on #151 i think this may be a better way to go.  if you agree i will tidyup and add any missing spec tests


xaque208:
This seems reasonable to me.

PR #153, opened 2018-01-26
add local_zone and domain_insecure parameters

b4ldr:
This is an update of #151 and #152


manifests/init.pp
```diff
@@ -6,6 +6,8 @@
   Hash $forward,
   Hash $stub,
   Hash $record,
+  Optional[Unbound::Local_zone] $local_zone,
```
xaque208: Just looking over the man page, looks like multiple local-zone definitions are allowed.  This should probably be `Optional[Array[Unbound::LocalZone]]` to allow multiple local zones.  The template will also need adjusting.

b4ldr: `type Unbound::Local_zone = Hash[String, Unbound::Local_zone_type]` so you can pass multiple zones here, template already accounts for this

xaque208: Yep, I just missed that.

xaque208:
This looks good.  Thanks for the contribution.

Issue #145, opened 2017-10-25
Options not found


helmerfud:
hello,
i'm new user of your module and i did not find log-queries and log-replies options.
it's a security concern in my enterprise.
is it possible to add these options ?

thank you.

xaque208:
Hi, thank you for the note.  Those options are absolutely welcome!  Is this something that you would be able to send a patch for?

b4ldr:
this has been merged so can be closed now

xaque208:
Thanks @b4ldr.

PR #155, opened 2018-02-12
add log options

b4ldr:
Hi, This merge request adds the options discussed in #145.  I have also tidied up the unbound.conf.erb file by adding a helper function to transfer bools into yes no strings.   i also add `log-time-ascii` which was in the manifest but not the template, i changed the default to false as that is the unbound default and as this value was not in the template that is the value that will be used today


xaque208:
This looks good to me.  Thanks for tidying up the template while you're there!

Issue #138, opened 2017-04-18
Should add support for ssl-upstream config option


xaque208:
Tunneled DNS looks to be overlooked in this module.  We should add support.

xaque208:
https://blog.apnic.net/2016/05/27/dns-privacy/

PR #156, opened 2018-02-14
138 tls options

b4ldr:
Ok so i started to look at adding options for #138 and got a bit carried away and have now ended up refactoring things significantly.  The main changes are
  * reorder the data/common.yaml and manifest/init.pp params so they are in the same order as the unbound documentation.  this makes it much easier to see what params are supported and which ones are not
  * Add types to all manifest params
  * require stdlib 4.13.0 so we can use `Stdlib::Absolutepath`
  * add helper function to make adding new parameters much easier
  * added ubuntu_version fact
  * use the above fact and only add parameters if the installed version of unbound supports them
 * add some additional types to make ease readability
 * add some white space to ease readability 
 * added support for almost all parameters in the unbound docs (ithink i only left out debug ones and ones that dont make much sense)
 * tried to add support for tags, however i dont use tags so dont know how they work and therefore not sure if i got it right (this last one i will factor out as a separate request)

So like i said i got a bit carried away :)

I would like to add some additional acceptance testing leveraging beaker so i can look at the open trust_anchor issue.  however before i  go down that route i wanted to give you a heads up of what i got so far.  I personally think most of the changes are cosmetic and would prefer to not have to spend loads of time splitting these up into seperate pulls, however i appreciate that it will probably need to be split into  at least a couple.  from my point of view i would prefer to spend my effort creating better rspec and beaker tests let me know your thoughts and thanks again for the module


lib/facter/unbound_version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+unbound_bin = '/usr/sbin/unbound'
```
xaque208: Hmm, my FreeBSD servers might fail with this path.

b4ldr: are you sure my bsd server is a bit behind but

```
ssh% uname -a
FreeBSD ssh.example.org 11.0-RELEASE-p15 FreeBSD 11.0-RELEASE-p15 #0: Tue Nov 14 08:04:40 UTC 2017     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
ssh% which unbound
/usr/sbin/unbound

Comment (xaque208) [reply lib/facter/unbound_version.rb]: I'm not certin.  This is what I was thinking though: https://github.com/xaque208/puppet-unbound/blob/master/data/os/FreeBSD.yaml#L3  Since we install the package with this module, we should probably be using the package's unbound.  I assume the `/usr/sbin/unbound` is from the unbound in base, which I think lacks TLS support.  Perhaps `which -a unbound`?

Comment (b4ldr) [reply lib/facter/unbound_version.rb]: yes you could be right, ill set up some beaker tests to  validate all this.  obviouly the bsd can not be done by travis unless you know something i dont but ill create some tests eaither way

**Comment by xaque208** [comment]
This all looks pretty good, and the tests are passing with only a few changes.  The logic for detecting the unbound version using the path probably should be corrected here.  As for the ordering, I'm not sure that using the manual is the right source for order, as it might be hard to guarantee in the future.  I'm not worried about the classes, but perhaps in the hiera yaml files, I tend to `!sort` in my vim.  It'd also break at first PR if I don't check the man page first, etc.

**Comment by b4ldr** [comment]
In relation to ordering the reason i choose the man page is so that i could easily scan down the man page with with the class definition open and quickly see which params where missing.  As such i would still like to keep that order in place in at least the class.  As to the hiera i think it might be best for you to accept this PR (after other fixes) and the re-sort how you like. 

how does that sound, ill look at the path thing later today

File: lib/facter/unbound_version.rb (hunk 1 above)
Comment (xaque208) [reply lib/facter/unbound_version.rb]: Personally I've not used beaker much, but test on long lived installations.  I think some unit tests for the platform would be fine, but however you want to approach that is fine.  I  still need to get openbsd updated in the facterdb code, which I've been meaning to do for months now.

**Comment by xaque208** [comment]
That all sounds good.  Thanks for the efforts @b4ldr.

**Comment by b4ldr** [comment]
ok i fixed the fact path the only other thing im unsure about is the `access-control` param im not really sure how that works and just to my best guess but may have overcomplicated things.  do you want me to strip this out and send a second PR for review?

https://github.com/xaque208/puppet-unbound/pull/156/files#diff-60ae41fd0a31977447947f59940ee9a4R58
https://github.com/xaque208/puppet-unbound/pull/156/files#diff-60ae41fd0a31977447947f59940ee9a4R58
https://github.com/xaque208/puppet-unbound/pull/156/files#diff-eadf16f413d37fbebbf52781d55a866a

File: manifests/init.pp [hunk 2]
```diff
@@ -3,87 +3,159 @@
...
+  Array[String]                                        $interface,
+  Boolean                                              $interface_automatic,
+  Optional[Array[String]]                              $outgoing_interface,           # version 1.5.10
+  Optional[Integer[1]]                                 $outgoing_range,
+  Unbound::Range                                       $outgoing_port_permit,
+  Unbound::Range                                       $outgoing_port_avoid,
+  Boolean                                              $outgoing_port_permit_first,
+  Optional[Integer[0]]                                 $outgoing_num_tcp,
+  Optional[Integer[0]]                                 $incoming_num_tcp,
+  Integer[0,4096]                                      $edns_buffer_size,
```
igalic: i'm curious to know where this limit comes from

templates/unbound.conf.erb
```diff
@@ -1,162 +1,217 @@
 # Managed by Puppet
 #
 <%-
-  def bool2yesno (value)
-    return value ? 'yes' : 'no'
+  if @unbound_version
+    @_unbound_version = @unbound_version
```
igalic: `_unbound_version`, doesn't need to be a class static variable, so you can leave the `@` out

templates/unbound.conf.erb
```diff
@@ -1,162 +1,217 @@
...
+  else
+    # this happens if unbuond is not installed i.e. first run
+    @_unbound_version = '0.a'
+  end
+  def print_config(name, value, version=false )
```
igalic: both of these functions could be converted to, well, functions.

manifests/init.pp
b4ldr: in theory the ends buffer size can be 64k as per the spec however 4k is generally the limit rfc 6891 hits at this (https://tools.ietf.org/html/rfc6891#section-6.2.5) and i believe nsd, unbound,  bind, knot etc all have this value hard coded in the source code (but its been a while since i checked)

templates/unbound.conf.erb
b4ldr: thanks for the review i addressed the other comments but im not sure what you meant by this one.

manifests/init.pp
igalic: :+1:

xaque208:
The access control bits you put in seem fine to me.  What part are you worried about?  Does it work?

xaque208:
Yep, I agree.  Thanks for the PR @b4ldr.  This looks good for merge.

PR #157, opened 2018-02-21
Added option to use ip-freebind.

tmichelis:
Hej,

we have to use ip-freebind instead of interface-automatic as it is still marked as experimental. We would like to use this module from puppet-forge instead of maintaining it by ourselves. That's why I ask for this pull request.

Cheers
Tobias

### DIFF
--- a/templates/unbound.conf.erb
+++ b/templates/unbound.conf.erb
@@ -21,7 +21,7 @@ server:
 <% end -%>
   log-time-ascii: <%= bool2yesno(@log_time_ascii) %>
   log-queries: <%= bool2yesno(@log_queries) %>
-  log-replies: <%= bool2yesno(@log_replies) %>
+  #log-replies: <%= bool2yesno(@log_replies) %>
   extended-statistics: <%= bool2yesno(@extended_statistics) %>
 <% if @statistics_interval -%>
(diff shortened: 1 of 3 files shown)


xaque208:
HI @tmichelis, Thanks for the PR.  Can you check the latest code to see if this is still required?  This will need a rebase at least due to conflicts that were just merged in.

tmichelis:
Yes I will. I will come back to you. Thanks you

b4ldr:
This was added in https://github.com/xaque208/puppet-unbound/pull/156 so you should be good now

templates/unbound.conf.erb
```diff
@@ -21,7 +21,7 @@ server:
 <% end -%>
   log-time-ascii: <%= bool2yesno(@log_time_ascii) %>
   log-queries: <%= bool2yesno(@log_queries) %>
-  log-replies: <%= bool2yesno(@log_replies) %>
+  #log-replies: <%= bool2yesno(@log_replies) %>
```
b4ldr: why comment this out, did you see an issue?

xaque208:
@tmichelis Hi.  Are you able to come back to this work?

PR #158, opened 2018-02-23
acceptance testing

b4ldr:
a couple of small fixes and some acceptance tests

I notice you have your own modulsync config so some of theses changes may have to go over there but wanted to create thee PR so i could test some the new test


xaque208:
This all looks reasonable, but I know almost nothing about acceptance testing in Puppet.  How does one run the tests?  Since most of my modules are BSD focused, and travis doesn't support BSD, I'd thought it fruitless.

b4ldr:
@xaque208 you can look at the travis-ci files to see how to run the acceptance tests e.g. for ubuntu

`BEAKER_set=docker/ubuntu-14.04 bundle exec rake beaker`

as you say travis-ci only supports docker so you cant run theses tests however you can still run them from the cli using vagrant and virtual box with the following

```
 BEAKER_destroy=no BEAKER_set=freebsd-10.3 bundle exec rake beaker
 BEAKER_destroy=no BEAKER_set=freebsd-11.1 bundle exec rake beaker
```

spec/acceptance/unbound_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'unbound class' do
+  describe 'running puppet code' do
+    it 'work with no errors' do
+      pp = 'class {\'::unbound\': }'
```
igalic: might wanna just use `"class {'::unbound': }"`, and, since we're testing >=4, we can also drop the `::` and go with `"class {'unbound': }"`

spec/acceptance/unbound_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+    describe service('unbound') do
+      it { is_expected.to be_running }
+    end
+    describe port(53) do
+      it { is_expected.to be_listening }
+    end
+    describe command('dig +dnssec . soa @localhost') do
+      its(:stdout) { is_expected.to match %r{\.\s+\d+\s+IN\s+SOA\s+a\.root-servers\.net\.\snstld\.verisign-grs\.com\.\s\d+\s1800\s900\s604800\s86400} }
```
igalic: this looks unreadable and i wish there was a better way :joy:

b4ldr: I agree, if you think of one ;) ...

PR #160, opened 2018-04-06
Update to hiera v5

jlutran:
This PR updates hiera.yaml to version 5 and removes the following warnings when running puppet 4.9+ :

Warning: Use of 'hiera.yaml' version 4 is deprecated. It should be converted to version 5
Warning: Defining "data_provider": "hiera" in metadata.json is deprecated.

### DIFF
--- a/hiera.yaml
+++ b/hiera.yaml
@@ -1,14 +1,13 @@
 ---
-version: 4
-datadir: data
+version: 5
+defaults:
+  datadir: 'data'
+  data_hash: 'yaml_data'
 hierarchy:
   - name: "release"
-    path: "os/%{facts.os.family}/%{facts.os.release.major}"
-    backend: yaml
+    path: "os/%{facts.os.family}/%{facts.os.release.major}.yaml"
   - name: "family"
-    path: "os/%{facts.os.family}"
-    backend: yaml
+    path: "os/%{facts.os.family}.yaml"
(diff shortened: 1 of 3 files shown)


xaque208:
Hi @jlutran, thanks for the PR.  This looks good to me.  The acceptance tests are new to this module, so I'm not sure why they are failing.

@b4ldr What do you make of the acceptance tests here?  Just a transient failure?  If so, is there a way that we can make the tests more deterministic?

xaque208:
Thanks for the comment @b4ldr.  I'll have a look over the modulesync configs for the failure allowance on the acceptance jobs.

b4ldr:
personally i don't think you should add them to allow_failures but thats your call.  As to the modsync changes i noticed the branch you have is a bit out of sync with the voxpupuli repo which has a few additional options for adding theses acceptance tests.  let me know if you need a hand

xaque208:
I tend to agree with that sentiment @b4ldr.  I've been caught up in non-puppet for a while.  I'll have a look at the changes you mention.  Though, I will say that not all the changes from voxpupuli are desirable sometimes.

In the mean time @jlutran, if we can bump the puppet agent version in `spec/spec_helper_acceptance.rb` to get the tests to pass, I'll merge.

xaque208:
Thanks for the PR @jlutran.

PR #161, opened 2018-04-09
Allow to set parameter forward-ssl-upstream for forward zones in

buzzdeee:
order to enable DNS via TLS.

This allows me to configure DNS via TLS for upstream resolvers i.e quad9 or cloudflare as well as having nsd running for local stub zones.

The test that tested for forward-first was enhanced to also include parameter forward_ssl_upstream


xaque208:
HI @buzzdeee!  Glad to see you around these parts. :)  Thanks for the PR.

PR #168, opened 2018-05-09
Fix regular expression in fact `unbound_version`.

denzery:
`\d` matches a single digit, so a version like `1.4.20` will result in the fact `unbound_version` containing `1.4.2` instead of `1.4.20`. The new expression will match a number (one or more digits), followed by any number of dots which are immediately followed by a number. This way we're able to match `1`, `1.2`, `1.2.3`, `1.2.3.4` etc.

### DIFF
--- a/lib/facter/unbound_version.rb
+++ b/lib/facter/unbound_version.rb
@@ -9,7 +9,7 @@ unbound_bin = case Facter.value('kernel')
 if File.exist? unbound_bin
   unbound_version = Facter::Util::Resolution.exec(
     "#{unbound_bin} -V 2>&1"
-  ).match(%r{\nVersion\s(\d\.\d\.\d)})[1]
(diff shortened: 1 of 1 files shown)


xaque208:
Thanks for the PR.  Would you be able to add a test for this change?  The regex looks good, but a test would make sure this supports the rest of the version strings you've mentioned above.

xaque208:
I've tried to add a test in #173 but it didn't work at first pass.  I can come back to look closer later.

xaque208:
#207

xaque208:
Thank you for the PR here.  Sorry for such a slow response.  I've tried three times now at different time s to get a test that works for this and have failed every time.  I've tested this manually and all looks good to me.  I'll merge https://github.com/xaque208/puppet-unbound/pull/211 to remove the newline.

PR #169, opened 2018-05-17
Fix regressions

fklajn-opera:
This pull requests fixes two bugs:
- typo in Unbound::Size pattern regex
- missing msg-cache-size parameter in the config template


xaque208:
Thanks for the PR.

Issue #166, opened 2018-04-20
interface changes don't take affect


b4ldr:
If you change the `$interface` while you have `$control_enable` the interface change does not take affect.  This is because if control_enable is true the service refreshes using `unbound-control reload`.  however as per the documentation

> The  interfaces  are not changed on a reload (kill -HUP)  but only on restart. 

This means that any changes to the interfaces (even if it is the first run) do not take affect until a manual restart is preformed.  The simple fix would be to remove the offending block[1] however this would mean a full restart of the daemon for any config changes which may not be desirable.  A more hacky way would be to store the interfaces in a file and restart the service if that file changes.  something like

```puppet
file {"$confdir/interfaces.txt":
  ensure => file,
  notify  => Exec['etc/init/.d/unbound restart'],
  content => $interfaces.join('\n'),
}
exec {'/etc/init/.d/unbound restart':
  refreshonly => true,
}
 ```
what do you think?
[1]https://github.com/xaque208/puppet-unbound/blob/master/manifests/init.pp#L189-L192

xaque208:
I believe this is complete.

PR #170, opened 2018-05-17
add method to restart unbound on interface change

b4ldr:
pull request to implement #166

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -155,6 +155,7 @@ class unbound (
   String                                               $service_name,
   Boolean                                              $service_hasstatus,
   String                                               $validate_cmd,
+  String                                               $restart_cmd,
   Array[String]                                        $custom_server_conf,
   Boolean                                              $skip_roothints_download,
 ) {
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -186,6 +186,14 @@
   }
 
   if $control_enable {
+    file {"${confdir}/interfaces.txt":
+      ensure  => file,
+      notify  => Exec['/etc/init/.d/unbound restart'],
+      content => "# Used by puppet-unbound\n${interface.join('\n')}",
+    }
+    exec {'/etc/init/.d/unbound restart':
```
xaque208: This restart command should be tunable I think.

b4ldr:
@xaque208 agreed and updated, i think this default should work for bsd, ubuntu and debian

xaque208:
Hmm, this isn't quite right.  I'm guessing its probably `systemctl` most everywhere on linux by now, no?  OpenBSD would be `/usr/sbin/rcctl restart unbound` and FreeBSD would be `/usr/sbin/service unbound restart`  I can add those before commit if helpful.

b4ldr:
both ubuntu and redhat use service as a wrapper top either systemctl, /etc/init.d or what ever upstart uses.  ill add something in for openbsd

b4ldr:
i have added exceptions for openbsd and freebsd.  I have also moved to using systemctl by default as it looks like the `service` command is a back-port hack which is likely gonna be removed in the future.  Further i changed the default of redhat 6 and ubuntu 14.04 to use the service command because i know that didn't have system.d.  i suspect there will need to some more exceptions but i dontknow them

xaque208:
Runs great in my environment.  Thanks @b4ldr.

PR #173, opened 2018-05-21
Fix facter regex test

xaque208:


spec/unit/facter/unbound_version_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Facter::Util::Fact do
+  before { Facter.clear }
+
+  context 'unbound' do
+    context 'With version output' do
+      before do
+        expect(Facter::Util::Resolution).to receive('/usr/sbin/unbound').with('-V 2>&1').once { "\nVersion 1.5.10\n" }
+      end
+      it { expect(Facter.value(:unbound_version)).to eq('1.8.10') }
```
b4ldr: shouldn't you use the same version strings in the expect and before e.g. 
```ruby
it { expect(Facter.value(:unbound_version)).to eq('1.5.10') }
```

xaque208: Ah yes, I seem to remember it failing for something else.  We'll see what happens now.

PR #174, opened 2018-05-22
Add missing module-config parameter to the template

fklajn-opera:
module-config parameter removed in commit 8ab7ee7cdb7ec940c6f2fabd5c3aadca5a1448e4 is added again

### DIFF
--- a/templates/unbound.conf.erb
+++ b/templates/unbound.conf.erb
@@ -145,6 +145,7 @@ server:
 <%= print_config('caps-whitlist', @caps_whitlist) -%>
 <%= print_config('qname-minimisation', @qname_minimisation, '1.5.7') -%>
 <%= print_config('qname-minimisation-strict', @qname_minimisation_strict, '1.6.0') -%>
+<%= print_config('module-config', @module_config) -%>
 <%= print_config('private-address', @private_address) -%>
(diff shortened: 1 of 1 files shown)


b4ldr:
@fklajn-opera i wanted to make the module_config option a bit more complete which was why it was removed, however i forgot to actually add it back.  I have now created a new pull request https://github.com/xaque208/puppet-unbound/pull/176 can you take a look at that and comment if it works for your use case

b4ldr:
yes mine is a breaking change, i wonder how many people use this feature,  would be nice to not have to support strings.  will see what @xaque208 says.

out of curiosity how do you use this feature, the current module doesn't allow you to set any parameters for the dns64, cachedb, ipsecmod and python modules unless you use `custom_server_conf`.  are you adding adding config using `custom_server_conf` or are you using this to disable the validator?

fklajn-opera:
In my case this is used only to disable validator, the service resolves internal domains.

b4ldr:
ok thanks

xaque208:
We can take both of these PRs.  @b4ldr If you are good with a merge of this and cutting a point release for it, you can rebase your work  and we'll cut a minor for the breaking change.  Good?

b4ldr:
@xaque208 yes, sounds good to me

PR #181, opened 2018-07-04
Change default values for harden_short_bufsize & harden_large_queries

jlutran:
As per [unbound.conf](https://www.unbound.net/documentation/unbound.conf.html) documentation, `harden_short_bufsize`and `harden_large_queries`default value is `no`.


xaque208:
This looks good.  Thanks for the contribution.

Issue #177, opened 2018-05-24
Unable to call unbound::local_zone class anymore


jlutran:
Since PR #153 unbound::local_zone class does not exists anymore.
Unfortunately, my puppet code was heavily using this feature to configure different local zones regarding of host type, eg :

```
ensure_resource('unbound::local_zone', '.', { 'type' => 'refuse' })
if( $zone1) { ensure_resource('unbound::local_zone', 'my.zone1.net', { 'type' => 'transparent' }) }
if( $zone2) { ensure_resource('unbound::local_zone', 'my.zone2.net', { 'type' => 'transparent' }) }
```

How can I use this feature again ? 
Merging all local zones a in single hash and pass it as a ::unbound class parameter seems kind of hard because of puppet immutable variable design

jlutran:
I use unbound::local_zone as a resolver whitelist on my hosts :

1. Set resolver to localhost only
2. Disable all zones resolving ({ '.' => 'refuse' })
3. Add custom zones using multiple unbound::local_zone calls in separate puppet profiles

Your code above is working, but hard to implement in my case because `::unbound::local_zone` class is called from multiple puppet manifests.

xaque208:
@b4ldr I assume then that the `unbound` class would then contain a `create_resources('unbound::local', $local_zone...)` or are you thinking something different?

@jlutran Would storing all of the local_zone data in the same hiera data object be enough for you?  I'm curious to learn more about why local_zone is in multiple manifests on the same node.  I assume just for organization purposes.  Is that right?

b4ldr:
>@b4ldr I assume then that the unbound class would then contain a create_resources('unbound::local', $local_zone...) or are you thinking something different?

yes, it would mean going back to the old way of doing it.

xaque208:
Bringing back the local_zone class while also supporting a parameter on the unbound class that, if defined, is passed to create_resources seems like the best of both worlds.  I'm speculating here, as I don't use the feature, but users would be able to define the resources centrally in hiera, or scattered throughout the manifest.

jlutran:
> @jlutran Would storing all of the local_zone data in the same hiera data object be enough for you? I'm curious to learn more about why local_zone is in multiple manifests on the same node. I assume just for organization purposes. Is that right?

I use a default unbound profile, and then set some local zones on custom roles. For example : 

```
class profiles::unbound(
    $default_nameservers = hiera('public_nameservers', []),
[...]

PR #182, opened 2018-07-04
Restore unbound::local_zone function

jlutran:
This PR fixes issue #177

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,9 +188,9 @@ class {'unbound:'
 }
 ```
 
-Or, using unbound::local_zone
+Or, using unbound::localzone
 ```puppet
-unbound::local_zone { '10.0.10.in-addr.arpa.':
+unbound::localzone { '10.0.10.in-addr.arpa.':
   type => 'nodefault'
 }
 ```
--- a/README.md
+++ b/README.md
@@ -188,6 +188,13 @@ class {'unbound:'
 }
 ```
 
+Or, using unbound::local_zone
+```puppet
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -188,6 +188,13 @@ class {'unbound:'
 }
 ```
 
+Or, using the legacy way
```
xaque208: It doesn't have to be legacy.

manifests/local_zone.pp
```diff
@@ -0,0 +1,27 @@
...
+# == Parameters:
+#
+# [*local_zone*]
+#   (required) Unbound::Local_zone custom type
+#
+# [*config_file*]
+#   (optional) name of configuration file
+#
+define unbound::local_zone (
+  Unbound::Local_zone $local_zone,
```
xaque208: I think the type here needs to be something other than the name of the class.

xaque208:
Thanks for sending this in @jlutran, and apologies for causing trouble in your environment.  We might need to rename either the type or the defined type for this to work as is.

jlutran:
Ok, I changed ```unbound::local_zone```function to use ```Unbound::Local_zone_type``` type only.
This means we won't be able to pass a ```Unbound::Local_zone_type``` hash to configure multiple local zones at the same time but I don't think it's a big deal since use can use the new ```local_zone``` parameter from ```unbound``` class to do it now.

xaque208:
This looks good to me.  Thanks for the contribution.

Issue #183, opened 2018-07-14
Binding to 0.0.0.0 is a bad practice 


akondasif:
Greetings,

I am a security researcher, who is looking for security smells in Puppet scripts. I noticed instances of binding to 0.0.0.0. Binding an address to 0.0.0.0 indicates allowing connections from all IP addresses. I would like to draw attention to these instances. Binding to 0.0.0.0 may lead to denial of service attacks. Practitioners have reported how binding to 0.0.0.0 facilitated security issues for MySQL (https://serversforhackers.com/c/mysql-network-security), Memcached (https://news.ycombinator.com/item?id=16493480), and Kibana (https://www.elastic.co/guide/en/kibana/5.0/breaking-changes-5.0.html).    

I suggest to use a dedicated IP address other than 0.0.0.0. 

Any feedback is appreciated.  

Source: https://github.com/xaque208/puppet-unbound/blob/master/examples/init.pp

b4ldr:
@xaque208 below is what the unbound man page states so interfaces should probably be set to `['::1', '127.0.0.1']` however that is likely to break some setups

```
 Interface  to  use  to connect to the network. This interface is
              listened to for queries from clients, and answers to clients are
              given  from  it.  Can be given multiple times to work on several
              interfaces. If none are given the default is to listen to local-
              host.   The  interfaces  are not changed on a reload (kill -HUP)
              but only on restart.  A port number can be specified with  @port
              (without spaces between interface and port number), if not spec-
              ified the default port (from port) is used.
```

xaque208:
Thank you for sending the note.

I think its just the example that could be updated here.  We also require the user to pass in the listen address, as well as the access list that will be allowed query, but the user is expected to know the impact of the values they enter since its entirely site specific.

Its worth noting, binding does not alone allow connections.  Other security mechanisms that are beyond the scope of this module might be in place, such as packet filtering.  Also, the service itself has a notion of providing access from particular prefixes, so merely binding to an address is not necessarily sufficient to provide access.  Also, due to the often public nature of DNS, I don't really see this as bad practice or an issue generally, but perhaps I'm overlooking something.  My personal expectation of deployment of DNS infrastructure is that the security decisions around DNS are weighed to determine the impact to the particular deployment, which is quite specific to the need of that infrastructure.

[...]

PR #185, opened 2018-07-21
Change default interface value

xaque208:
In order to follow the unbound default, here we cease to require that an
interface is passed to the module.  Instead, we allow this parameter to be
optional.  When set, this parameter behaves just as it did before.  When the
user does not provide this value, the defaults are no longer passed in, and thus
absent from the configuration file on disk resulting in the default values
coming from unbound.conf(5) which is to listen on localhost.


xaque208:
#183

PR #186, opened 2018-07-27
Fix module-config

silkeh:
This adds a missing `"`


xaque208:
Yeah, I'm not sure what's up with the tests lately.  This looks good though.  Thanks for the PR.

Issue #184, opened 2018-07-17
Add SmartOS support


joelgarboden:
I created `data\os\Solaris.yaml` locally and got it working, tossing stub here for others to reference.

```
---
unbound::confdir: '/opt/local/etc/unbound'
unbound::pidfile: '/usr/local/etc/unbound/unbound.pid'
unbound::logdir: '/var/log/unbound'
unbound::fetch_client: 'wget -O'
unbound::control_setup_path: '/opt/local/sbin/unbound-control-setup'
unbound::control_path: '/opt/local/sbin/unbound-control'
unbound::validate_cmd: '/opt/local/sbin/unbound-checkconf %'
unbound::restart_cmd: "/usr/sbin/svcadm restart %{hiera('unbound::service_name')}"
unbound::anchor_fetch_command: "/opt/local/sbin/unbound-anchor -a %{hiera('unbound::auto_trust_anchor_file')}"
```

The hiera used `path: "os/%{facts.os.family}/%{facts.os.release.major}.yaml"` doesn't quite fit SmartOS' structure of `{"name"=>"SmartOS", "family"=>"Solaris", "release"=>{"minor"=>"11", "full"=>"5.11"}}` but using `Solaris.yaml` got me functional.

xaque208:
The following line of `path: "os/%{facts.os.family}.yaml"` in the hiera.yaml picks it up.  Feel free to send a PR for this change.  Could extend the hiera if needed.  I have no way to confirm the values that you've got, but thanks for sending the note!

xaque208:
@joelgarboden Are you able to send a PR for this change?

joelgarboden:
Apologies for the delay, been heads down with Puppet 3->5 conversion at work.

I'll try to get a PR up this week.

PR #187, opened 2018-08-15
Add SmartOS support #184

joelgarboden:
Relates to #184, adds SmartOS support.

hiera.yaml tweak was required to differentiate between Solaris and SmartOS.


xaque208:
Great, thank you @joelgarboden.  The test failures here look unrelated.

xaque208:
From the failing test here, looks like we might not be pinned on a gem perhaps?  @b4ldr  This will likely come up in future PRs.

PR #189, opened 2019-02-26
Update from xaque208 modulesync_config

xaque208:


xaque208:
Non-beaker tests passing, so merging.  I'll come back for docker stuff.

PR #193, opened 2019-04-19
Support TLS

xaque208:


xaque208:
Passing enough.

anthonysomerset:
a heads up....

this breaks things on Ubuntu 18.04 because the default unbound version is 1.6.7 and tls-upstream comes in 1.7.0, some checks might be needed to no include it when package version is below 1.7.0 or on specific systems

xaque208:
@anthonysomerset Thanks for the heads up.  Are you able to test the `tlsversion` branch to confirm the fix?

PR #198, opened 2019-06-06
Set harden-referral-path to false by default.

jensalmer:
This feature is labeled experimental, burdens authority servers and is
not RFC standard. I have personally found problems with this option
being enabled.


xaque208:
Looks like the default is no also.  Thanks for the PR.  Failing tests here are just beaker.

PR #200, opened 2019-07-08
Issue 199: add version checking for ssl/tls parameters

### LINKED ISSUE #199: version 2.4.3 breaks the configfile for tls-upstream on CentOS 7
The unbound version in the CentOS supported repo's is too old to support tls-upstream and other tls-* keywords in the configfile. Those keywords where added  in module version 2.4.3 as a future replacement for the ssl-upstream (etc) versions.  We should skip those on CentOS 7 (and also RedHat 7 and below)

b4ldr:
As far as i can tell theses parameters where added in 1.7.0 

fixes #199


xaque208:
Just beaker broken here.  Thanks for the PR @b4ldr.

PR #201, opened 2019-07-10
beaker: fix beaker

b4ldr:
im not sure why but to use my ed25519 key i needed to use rbnacl <5 but can remove this if you want.  I also added a test for Debian 9.  I'll test buster and bionic once this is merged

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -26,3 +26,8 @@
       env: BEAKER_set=docker/debian-8 CHECK=beaker
       services: docker
       sudo: required
+    - rvm: 2.3
+      bundler_args: --without development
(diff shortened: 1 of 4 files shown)


xaque208:
Hooray!  Thanks @b4ldr for coming back to this.  Beaker remains a mystery to me.

b4ldr:
no probs feel free to ping me if it breaks i try to keep an eye on things but can obviously missed this.  also looks like litmus[1] may be replacing beaker [ish] and the tests in this module should be compatible with that.

[1]https://puppet.com/blog/litmus-new-module-acceptance-testing-tool

xaque208:
Ah, interesting.  Probably something that will need to get worked into the modulesync bits.

Issue #180, opened 2018-05-30
pid dir permissions could cause problems


b4ldr:
the unbound manifest tries to set the permissions of the pid file  base directory[1].  for debian this is set to `/run/unbound.pid`[2].  As such puppet tries to set the dir permissions for `/run` to `unbound` which is not desirable. for now im going to just.  Im not sure if this is a bug or desirable, especially if we consider chroot environments.  for now i have just set `unbound::pidfile: ~` however i did need to patch the module[3] to get that to work.  also note that `dirname` doesn't act as i expected when pidfile was blank[4]

[1]https://github.com/xaque208/puppet-unbound/blob/master/manifests/init.pp#L178-L180
[2]https://github.com/xaque208/puppet-unbound/blob/master/data/os/Debian.yaml#L2
[3]https://github.com/xaque208/puppet-unbound/pull/179
[4]https://github.com/puppetlabs/puppetlabs-stdlib/pull/913

xaque208:
Maybe the permissions for the pid dir could only get set if the directory name ends with `unbound`?  Just a thought.

b4ldr:
fixes for this got merged

https://github.com/xaque208/puppet-unbound/pull/179

also worth pointing out that the fixes to stdlib basedir have also been merged.  can you cut a new release.

Cheers

xaque208:
Yep, can do.

chrisboulton:
I'm not sure #179 is the right fix for this. Regardless of the changes in it, the module by default (on Debian) is still changing the owner of `/run` to `unbound` which is definitely undesirable.

I'm not sure the suggestion in the comments earlier is great. It makes a lot of assumptions.

Is the only distro that actually requires something like this (at least out of the box -- where the run directory isn't an OS provided one), FreeBSD? That's all I can spot from the data files.

xaque208:
I think freebsd creates the directory as part of the port.  I can check later.  OpenBSD ships with unbound by default, as does FreeBSD, so I'd assume the directories needed for the base unbound would not be an issue.  We could do something like not manage the permissions, but then we're not able to guarantee that the daemon can read the files necessary.

@chrisboulton Is there another action you're suggesting we take here?

b4ldr:
@chrisboulton first please send code if you have it.    

Now I am unsure as to what assumptions you think i made.  however let me try to address your comments

>Regardless of the changes in it, the module by default (on Debian) is still changing the owner of /run to unbound

[...]

PR #202, opened 2019-07-12
attempt a patch to 180

b4ldr:
rebase of: https://github.com/xaque208/puppet-unbound/pull/190

i'm still not sure if this is the best idea but i thought i would rebase, also i'm no longer with icann

original msg:
its not pretty and i hate the variable name but something to think about for #180


xaque208:
Closing in favor of #210

PR #204, opened 2019-08-23
Add service control and package ensure parameters

cohoe:
This PR adds parameters for `$service_enable` and `$service_ensure` to the module. There are cases where I need the ability to stop/disable unbound.

Update: also adds parameter for `$package_ensure` to enable the ability to pin a specific version (default is `installed` aka `present` which maintains compatibility.

The `Enum[]` for `service_ensure` was lifted from [puppetlabs-ntp](https://github.com/puppetlabs/puppetlabs-ntp/blob/master/manifests/init.pp#L261). Per the [Puppet Docs](https://puppet.com/docs/puppet/4.6/types/service.html#service-attribute-ensure) (4.6 to latest) this should be able to accept values of `stopped`, `running`, `false`, or `true`. However the [Enum Data Type](https://puppet.com/docs/puppet/latest/lang_data_abstract.html#the-enum-data-type) can only match strings, thus rendering the boolean values unavailable. Given the Puppetlabs reference module has this behavior as well, I suspect this will not be a problem for the majority of users who elect to use this parameter here.


cohoe:
Tests failed due to what I believe are unrelated OpenBSD configurations.

```
  1) unbound on openbsd-5.9-amd64 with default params should contain Package[unbound]
     Failure/Error: it { is_expected.to contain_package(package) }
       expected that the catalogue would contain Package[unbound]
```
Package name is set to [emptystring ](https://github.com/xaque208/puppet-unbound/blob/master/data/os/OpenBSD.yaml#L9) which prevents the `package{}` resource from being declared in the catalog.

```
2) unbound on openbsd-5.9-amd64 with default params should contain Concat::Fragment[unbound-header] with content =~ /\s+root-hints:\s"\/var\/unbound\/etc\/root.hints"/, content =~ ........
     Failure/Error:
       is_expected.to contain_concat__fragment(
         'unbound-header'
       ).with_content(........
       expected that the catalogue would contain Concat::Fragment[unbound-header] with content set to /\s+pidfile:\s"\/var\/run\/unbound\/unbound.pid"/ but it is set to "# Managed by Puppet\n#......
```

pidfile is set to `/var/run/unbound.pid` in [hiera](https://github.com/xaque208/puppet-unbound/blob/master/data/os/OpenBSD.yaml#L3) but `/var/run/unbound/unbound.pid` in [test](https://github.com/xaque208/puppet-unbound/blob/master/spec/classes/init_spec.rb#L23). I do not operate any OpenBSD systems so I do not know which one is appropriate.

xaque208:
Thank you for the PR.  I can take a look at the OpenBSD issue over the weekend and come back to this.

PR #205, opened 2019-08-26
Fix neg-cache-size in unbound.conf

cohoe:
The config key of `neg-cache-slabs` does not exist in [unbound.conf(5)](https://nlnetlabs.nl/documentation/unbound/unbound.conf/), and the service fails to start if it is present. I believe this to be a typo and that the key of `neg-cache-size` should be used, especially since that is the parameter that is being passed in.

```
Error: Execution of '/usr/sbin/unbound-checkconf /etc/unbound/unbound.conf20190826-27124-6oiy2x' returned 1: /etc/unbound/unbound.conf20190826-27124-6oiy2x:83: error: unknown keyword 'neg-cache-slabs'
```

This comes from unbound-1.6.6.1-el7.x86_64 and an internal repackage of 1.5.9.1-el6.x86_64


xaque208:
Great, thank you for the fix.

cohoe:
Appreciate your time and code!

PR #210, opened 2019-09-29
stop managing system directories like /run

tequeter:
This is similar to #202, but simpler.

I had to declare the newer Debian versions to exercise the unit tests on them.

(Nevermind these acceptance results on my comp, you have Travis set up - The acceptance tests run successfully on Debian 8 and 9. The puppet agent does not install on Debian 10 nor Ubuntu 18 and I didn't bother to debug it. The tests fail on CentOS 7 because there is no `dnsutils` package available.)

### DIFF
--- a/data/os/Debian/6.yaml
+++ b/data/os/Debian/6.yaml
@@ -1,0 +1,2 @@
+---
+unbound::pidfile: '/var/run/unbound.pid'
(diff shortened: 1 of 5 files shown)


xaque208:
These look like good changes to me.  Thank you for putting this up for review.

xaque208:
Thanks for the contribution.

PR #212, opened 2019-10-07
unbound_version: add spec tests to the unbound_version fact

b4ldr:
@xaque208 here are some tests to replace the ones attempted in #173


xaque208:
The test errors here look similar to what I was running into.

b4ldr:
oh i forgot to fix rubocop errors should be god now

b4ldr:
strange this works on my laptop but not a fresh install, suspect something is missing from the Gemfile but not sure what just yet

b4ldr:
Ok i think i got it, the fact was testing for the presence of the unbound binary.  if it doesn't exist the fact will return nil and it doesn't exist in travis so we have to mock that a bit.

PR #216, opened 2019-12-15
Add functionality to render local data and override local-zone template:

findmyname666:
- Add a new param local_data which is an Array of Strings into `localzone.pp`. local_data will be add into a configuration file as defined in a template.
  - Add a new param template_name into `localzone.pp`. Therefore anybody can use a custom template.


manifests/localzone.pp
```diff
@@ -9,23 +9,39 @@
...
+# [*template_name*]
+#   (optional) String.
+#   Use a custom template.
+#   Default value: 'unbound/local_zone.erb'
 #
 define unbound::localzone (
   Unbound::Local_zone_type $type,
   String $zone = $name,
   $config_file = $unbound::config_file,
+  Array[String] $local_data = [],
```
b4ldr: it might be nice to create a custome type for resource records instead of just accepting any random string.  something like
```puppet
type unbound::local_data =  Struct[{
    name => String,
    type => String,
    data => String,
  }]
]
```
in yaml you would have
```yaml
unbound::local_data:
  - name: foobar
    rrtype: A
    data: 192.0.2.1
  - name: foobar
    rrtype: TXT
    data: 'spf=foobar'
```

xaque208: This seems like a good suggestion to me.  It will slightly complicate the template, but not much.

findmyname666: Thanks for ideas. I adjusted PR.

xaque208:
This looks great.  Thank you for the PR.

PR #220, opened 2020-01-21
validate_unbond_addr: replace functionality with a custom type

b4ldr:
This PR removes the custom validate function and replaces it with type checking.  The patterns in the custom type are copied from `Stdlib::IP`.  Its not very DRY however i don't think the updated patterns are useful enough to add to stdlib and i don't know how to extend the patterns to achieve the functionality we desire as such this seemed the best comprimise


xaque208:
Looks good, thank you.

PR #222, opened 2020-01-25
Add test for unbound::stub

xaque208:
Without this change, there is no facility to know if changes to the
unbound::stub define cause issues on any of our platforms.  Here we ensure a
simple test catches the compile failures.


xaque208:
I see now that `stub-name` will need re-implementing.

Issue #223, opened 2020-02-07
Fix installation on Debian distribution - e.g. unbound option auto-trust-anchor-file is provided two times


findmyname666:
Unbound package on Debian provide two configuration file into path **/etc/unbound/unbound.conf**:
```
# cat /etc/unbound/unbound.conf.d/root-auto-trust-anchor-file.conf /etc/unbound/unbound.conf.d/qname-minimisation.conf 
server:
    # The following line will configure unbound to perform cryptographic
    # DNSSEC validation using the root trust anchor.
    auto-trust-anchor-file: "/var/lib/unbound/root.key"
server:
    # Send minimum amount of information to upstream servers to enhance
    # privacy. Only sends minimum required labels of the QNAME and sets
    # QTYPE to NS when possible.

    # See RFC 7816 "DNS Query Name Minimisation to Improve Privacy" for
    # details.

    qname-minimisation: yes
```
Configuration files are provided on Stretch and Buster code-names.

I create additional files in this folder therefore i include all configuration files from it. Furthermore it would be unclear for maintainers when some configuration files aren't use.

Unbound linter is failing that the option is provided two times.
```
Info: Computing checksum on file /etc/unbound/unbound.conf
Info: /Stage[main]/Unbound/Concat[/etc/unbound/unbound.conf]/File[/etc/unbound/unbound.conf]: Filebucketed /etc/unbound/unbound.conf to puppet with sum 7b9cf83ef566e394b3f259ae7b0efc7d
Error: Execution of '/usr/sbin/unbound-checkconf /etc/unbound/unbound.conf20200207-19370-19euxcc' returned 1: [1581070261] unbound-checkconf[21052:0] error: trust anchor presented twice
[1581070261] unbound-checkconf[21052:0] error: could not parse auto-trust-anchor-file /var/lib/unbound/root.key line 2
[1581070261] unbound-checkconf[21052:0] error: error reading auto-trust-anchor-file: /var/lib/unbound/root.key
[1581070261] unbound-checkconf[21052:0] error: validator: error in trustanchors config
[1581070261] unbound-checkconf[21052:0] error: validator: could not apply configuration settings.
[1581070261] unbound-checkconf[21052:0] fatal error: bad config for validator module
[...]

PR #224, opened 2020-02-07
Purge unmanaged configuration files

findmyname666:
There are added additional configuration files into **/etc/unbound/unbound.conf.d/** when package is installed from Debian mirror. This PR introduce logic to purge configuration files which aren't managed by Puppet and it provides a fix for #223.


findmyname666:
i will open another one.

PR #229, opened 2020-02-19
Fix order of parameters

dhoppe:
We should use the same order as in ```manifests/init.pp```.

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -1,7 +1,7 @@
 ---
 unbound::verbosity: 1
-unbound::statistics_cumulative: false
 unbound::statistics_interval: ~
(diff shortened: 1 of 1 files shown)


b4ldr:
i personally think unbounds default needs updating.  EDNS buffer size of 4096 causes issues with ipv6 and the [current RFCs](https://tools.ietf.org/html/rfc6891#section-6.2.3) and the [dns experts](https://github.com/dns-violations/dnsflagday/issues/125)  recommend using lower values between 1280 and 1400 unless you know fragmentation is not an issue.

That said unbound will try lower values if there is a failure so I'm not strongly against changing

PR #231, opened 2020-02-20
Fix on OS where root group != 'root'

buzzdeee:
There are plenty of OS out there, where group root doesn't exist,
so use 0 instead to refer to the root users group, i.e. might
be group wheel

This is the case for me on OpenBSD, and my usual fix when I 
encounter a group named 'root'


xaque208:
Yes, thank you.

PR #236, opened 2020-05-15
Restart for interface change after service is configured

nward:
When interfaces change, a restart is triggered. There is no ordering defined, so this can run before the configuration file, root keys, etc. are written - so the restart fails.

This PR solves this by adding some ordering to run the unbound restart command after the service resource has run - so anything which notifies or runs before the service resource will always run before the restart. This also means that the restart now runs *after* the service has started - so the initial config of the service means it will start and then restart. Given it's only the initial restart, I think that's ok.

I also change some titles for resources, to make testing easier - I have found that best practice is to define static titles for things like exec, and then make the command dynamic - so that they can be referenced easily in dependencies and specs.

I have been unable to run the specs - I have followed the instructions in the contributor guide, but it never seems to want to find any facts for the OSes. I am not at all familiar with the style of defining facts for tests in this module.
I did write tests - but am not sure if they are currently passing. The module works when run on real systems - I am using this.

Issue described above - happy to submit an issue if that is required for a PR.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -225,11 +225,13 @@ class unbound (
   if $control_enable {
     file {"${confdir}/interfaces.txt":
       ensure  => file,
-      notify  => Exec[$restart_cmd],
+      notify  => Exec['restart unbound'],
       content => template('unbound/interfaces.txt.erb'),
     }
-    exec {$restart_cmd:
+    exec { 'restart unbound':
+      command     => $restart_cmd,
       refreshonly => true,
+      require     => Service[$service_name],
     }
     Service[$service_name] {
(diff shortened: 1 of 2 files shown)


nward:
I have removed WIP - it seems the tests are doing on travis exactly what they were doing for me on my local machine.. Not sure I can do much there.

xaque208:
Thank you for the PR.  This looks reasonable me to me.  Looks like the failures are just rubocop complaining.  Its never satisfied.  But!  Try `rubocop -a` and maybe it will fix them for you.

xaque208:
Excellent, thank you.

PR #241, opened 2020-09-20
Add support for stub-no-cache option

xaque208:
Without this change, users are unable to modify the default stub option for
disabling the cache.  Here we include support for the option and a test.

### DIFF
--- a/manifests/stub.pp
+++ b/manifests/stub.pp
@@ -32,6 +32,7 @@ define unbound::stub (
   Variant[Array[Unbound::Address], Unbound::Address] $address,
   # lint:ignore:quoted_booleans
   Variant[Boolean, Enum['true', 'false']]            $insecure    = false,
+  Variant[Boolean, Enum['true', 'false']]            $no_cache    = false,
   # lint:endignore
   Unbound::Local_zone_type                           $type        = 'transparent',
   Stdlib::Unixpath                                   $config_file = lookup('unbound::config_file'),
--- a/templates/stub.erb
+++ b/templates/stub.erb
@@ -3,3 +3,6 @@ stub-zone:
 <% @address.each do |addr| -%>
   stub-addr: <%= addr %>
 <% end -%>
+<% if @no_cache == 'true' or @no_cache == true -%>
+  stub-no-cache: yes
+<% end -%>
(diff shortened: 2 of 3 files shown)


manifests/stub.pp
```diff
@@ -32,6 +32,7 @@
   Variant[Array[Unbound::Address], Unbound::Address] $address,
   # lint:ignore:quoted_booleans
   Variant[Boolean, Enum['true', 'false']]            $insecure    = false,
+  Variant[Boolean, Enum['true', 'false']]            $no_cache    = false,
```
b4ldr, line 35: Personally i would just make this Boolean, i think there was a need to make the other one a variant so as to make it backwards compatible but as this is a new param there is no need for that

templates/stub.erb
```diff
@@ -3,3 +3,6 @@ stub-zone:
 <% @address.each do |addr| -%>
   stub-addr: <%= addr %>
 <% end -%>
+<% if @no_cache -%>
```
b4ldr, line 6: if no_cache == 'false' this will pass as a none empty string is considered true

```
$ ruby -e "if 'false' then puts 'true' end "                                                               
true
```

b4ldr (changes requested):

manifests/stub.pp
igalic: i'd do that in future release, where we drop this type of backwards compatibility

templates/stub.erb
igalic: ya, but we won't get here with an empty string, given our type definition

igalic: oh, right, we don't normalise `'false'` to `false`

xaque208: Ah yes, clearly I've not been writing much ruby.  I'll fix.

manifests/stub.pp
xaque208: I'll keep the consistency here with the other param, but I agree that deprecating this to favor the boolean seems good longer term.

xaque208:
Just changed the template condition to be a bit more specific.  Thanks for the review

bastelfreak (approved):

manifests/stub.pp
bastelfreak: the next release will be a major one, so now might be the time to make it correct for the new param and fix the others in a separate PR?

PR #252, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (changes requested):
Not sure why it's failing on some systems with a timeout to `dnssec-failed.org`. Perhaps a temporary network issue?

b4ldr:
> Not sure why it's failing on some systems with a timeout to `dnssec-failed.org`. Perhaps a temporary network issue?

I cant replicate this issue locally but it seems to be pretty reproducible here.  As all the preseeding test use the same domain im not convinced its a temporary network issue.  Looks more like there may be some rate limiting either eiether in the container or at the dnssec-fail.org ns serveres.  Either way this looks good to me

b4ldr:
I reverted this as i hadn't noticed that the failures where in github but working in travis (now i fixed a [few](https://github.com/voxpupuli/puppet-unbound/pull/254) [issues](https://travis-ci.org/github/voxpupuli/puppet-unbound/jobs/750917936)).  could you regenerate this PR so i can investigate the github CI issues further

ekohl:
I opened https://github.com/voxpupuli/puppet-unbound/pull/257

PR #257, opened 2020-12-21
Modulesync 4.0.0

ekohl:
Since https://github.com/voxpupuli/puppet-unbound/pull/252 was reverted, here is a fresh version of it.


b4ldr:
See chat here https://github.com/voxpupuli/puppet-unbound/pull/258#discussion_r547346358 the ipv6 configuration is incorrect and needs to be first fixed in modsync_config (there are additional links to PR in that coment)

ekohl:
I've removed the IPv6 part here now so it should be green. I don't mind merging it even if it's not pure modulesync right now. It'll be a lot easier if everything is at least running on Github Actions.

b4ldr:
> I've removed the IPv6 part here now so it should be green. I don't mind merging it even if it's not pure modulesync right now. It'll be a lot easier if everything is at least running on Github Actions.

Ack sounds good to me will merge

PR #258, opened 2020-12-22
fix beaker tests

b4ldr:
DO NOT MERGE (this is just a to troubleshoot the workspace builds)
Seems that the beacker tests are sporadicly failing when running the
following command

 dig +dnssec dnssec-failed.org @localhost

Try fixing this by using 127.0.0.1 and +tcp

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: CI
+
+on: pull_request
+
+jobs:
+  acceptance:
+    runs-on: ubuntu-latest
+    env:
+      BUNDLE_WITHOUT: development:test:release
+    strategy:
+      fail-fast: false
+    name: 'Debian 10 puppet 6'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
+          bundler-cache: true
+      - name: Run tests
+        run: bundle exec rake beaker
+        env:
+          BEAKER_PUPPET_COLLECTION: puppet6
+          BEAKER_setfile: "debian10-64{hostname=debian10-64.example.com}"
--- a/spec/acceptance/unbound_spec.rb
+++ b/spec/acceptance/unbound_spec.rb
@@ -15,7 +15,7 @@ end
 describe 'unbound class' do
   describe 'running puppet code' do
     it 'work with no errors' do
-      pp = "class {'unbound': }"
+      pp = "class {'unbound': verbosity => 5, logfile => '/tmp/unbound.log'}"
       apply_manifest(pp, catch_failures: true)
       apply_manifest(pp, catch_failures: true)
       expect(apply_manifest(pp, catch_failures: true).exit_code).to eq 0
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,76 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        setfile: ${{fromJson(needs.setup_matrix.outputs.beaker_setfiles)}}
+        puppet: ${{fromJson(needs.setup_matrix.outputs.puppet_major_versions)}}
+    name: ${{ matrix.puppet.name }} - ${{ matrix.setfile.name }}
+    steps:
+      - name: Enable IPv6 on docker
+        run: |
+          echo '{"ipv6":true,"fixed-cidr-v6":"2001:db8:1::/64"}' | sudo tee /etc/docker/daemon.json
```
bastelfreak, line 63: My wild guess is that configuring a nonrouteable prefix might cause issues

b4ldr: I can test that however Im not sure its what is causing the issue.  The the tests are explicitly failing on the following (and only on some combinations of OS + puppet version)
`dig +tcp +dnssec SOA dnssec-failed.org @127.0.0.1`
but not:
`dig +dnssec +cd SOA dnssec-failed.org @127.0.0.1`

the nsset for dnssec-failed.org also has ipv4 and ipv6 nameservers so broken ipv6 shouldn't affect unbond in this way

b4ldr: The fact that the latter command works suggests that unbound is able to fetch all the data it needs to answer the former questions but cant for what ever reason

b4ldr: Ok turns out you where right, it seems to take a long time fetching all of the DNSSEC material as it needs to first get the DNSKEY/DS set from the the root, org and dnssec-failed.org zone.  As the data on dnssec-failed.org is bad by design it checks all 6 name-servers. Further because it thinks it has Ipv6 it has to first timeout on the ipv6 address before trying the v4 address. 

the subsequent query works within the time because we dont care about validation so unbound is just sending the bad data from the first thing to respond if its good or not.  i have created https://github.com/voxpupuli/modulesync_config/pull/694

spec/acceptance/unbound_spec.rb
```diff
@@ -15,7 +15,7 @@
 describe 'unbound class' do
   describe 'running puppet code' do
     it 'work with no errors' do
-      pp = "class {'unbound': }"
+      pp = "class {'unbound': verbosity => 5, logfile => '/tmp/unbound.log'}"
```
ekohl, line 18: If this happens more often, https://github.com/voxpupuli/voxpupuli-acceptance/#environment-variables-to-facts may be a good way to make this easier.

ekohl:
Given modulesync 4.0.0 was merged, I think this can be closed now.

spec/acceptance/unbound_spec.rb
b4ldr: this was mainly to debug the ipv6 issue and not sure if its genrally usefull but thanks for the pointer

PR #259, opened 2020-12-22
(#250) Add RPZ support

### LINKED ISSUE #250: Please support 'respip' in module_config
I would like to use the new rpz feature of unbound. however, in order to do so, I need to use the respip module, which is not in the list of allowed modules:

`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Unbound]: parameter 'module_config' index 0 expects a match for Unbound::Module = Enum['cachedb', 'dns64', 'ipsecmod', 'iterator', 'python', 'subnetcache', 'validator'], got 'respip' (file: /etc/puppetlabs/code/environments/ifi/modules/profile/manifests/unbound.pp, line: 165, column: 3)`

adding 'respip' to the array in types/module.pp solves the problem.

Best,
Hp

b4ldr:
#### Add support for RPZ
Fixes #250

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -211,6 +211,7 @@ class unbound (
   Stdlib::Absolutepath                          $unbound_conf_d                  = "${confdir}/unbound.conf.d",
   Unbound::Hints_file                           $hints_file                      = "${confdir}/root.hints",
   Optional[String[1]]                           $hints_file_content              = undef,
+  Hash[String[1], Unbound::Rpz]                 $rpzs                            = {},
 ) {
   $_base_dirs = [$confdir, $conf_d, $keys_d, $runtime_dir]
   $_piddir = if $pidfile { dirname($pidfile) } else { undef }
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -1074,6 +1074,83 @@ describe 'unbound' do
 
         it { is_expected.to compile.with_all_deps }
       end
+
+      context 'RPZs config' do
+        let(:params) do
+          {
+            module_config: ['respip'],
+            rpzs: {
+              'test1' => {
+                'primary' => ['192.0.1.2', 'primary.example.org'],
+              },
+              'test2' => {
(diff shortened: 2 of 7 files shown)


spec/classes/init_spec.rb
```diff
@@ -942,6 +942,47 @@
...
+            }
+          }
+        end
+        it { is_expected.to compile.with_all_deps }
+        it do
+          is_expected.to contain_concat__fragment(
+            'unbound-modules'
+          ).with_content(
+            %r{
+
```
b4ldr, line 981: This needs filling in

manifests/init.pp
```diff
@@ -211,6 +211,7 @@
   Stdlib::Absolutepath                          $unbound_conf_d                  = "${confdir}/unbound.conf.d",
   Unbound::Hints_file                           $hints_file                      = "${confdir}/root.hints",
   Optional[String[1]]                           $hints_file_content              = undef,
+  Hash[String, Unbound::Rpz]                    $rpzs                            = {},
```
bastelfreak, line 214: Does it make sense to allow empty strings? Otherwise we could do:
```suggestion
  Hash[String[1], Unbound::Rpz]                 $rpzs                            = {},
```

types/rpz/action.pp
```diff
@@ -0,0 +1 @@
+type Unbound::Rpz::Action = Enum['nxdomain', 'nodata', 'passthru', 'drop', 'disabled', 'cname']
```
bastelfreak, line 1: for both types, can you please add a `# @summary describe the type` and add a unit test? https://github.com/voxpupuli/puppet-unbound/tree/master/spec/type_aliases already has some examples.

b4ldr: i added types for the struct seemed a bit overkill for the enums but can if you insist

ekohl (approved):
I'm going to admit I didn't fully check the implementation. The code looks good to me and I trust you on the unbound specifics.

PR #261, opened 2021-03-29
custom_server_conf order

dsoltero:

### DIFF
--- a/templates/unbound.modules.conf.erb
+++ b/templates/unbound.modules.conf.erb
@@ -1,5 +1,3 @@
-# Managed by Puppet
-#
 <%-
   def unbound_version
     version = @unbound_version ? @unbound_version : '0.a'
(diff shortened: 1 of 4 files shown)


templates/unbound.modules.conf.erb
```diff
@@ -1,5 +1,3 @@
-# Managed by Puppet
```
ghoneycutt, line 1: This should probably stay.

ghoneycutt (changes requested):
This looks great! Please add a spec test that shows the content is as you would expect.

templates/unbound.modules.conf.erb
dsoltero: that is concat into the unbound.conf that already states that at the beginning . Basically that is a duplicate

PR #264, opened 2021-04-20
Refactor

b4ldr:

 * $package_name is not optional as such remove the logic which assumes it is optional and specify the dependencies in place
 * drop create_resources() in favour of using each
 * for `$dirs` moving to using each allowed us to simplify the ownership issue

No bug fixes just a clean up


b4ldr:
> $package_name is not optional as such remove the logic which assumes it is optional and specify the dependencies in place
This is not true OpenBSD has this set to a blank string.  we should make it Optional  will fix this shortly

kenyon (approved):

PR #265, opened 2021-04-20
move defaults to class so that puppet-strings can build better docs

b4ldr:

* I also updated the defaults for arrays to []
* I also updated the defaults for hash to {}
* Add some types to make init.pp a bit more readable

Currently puppet-strings is unable to read defaults from hiera as such move the defaults back to init.pp


b4ldr:
Do be honest i don't really like this patch, I think that puppet-strings should be able to read defaults from hiera, Ideally displaying the default for the various OS's.  As the [puppet labs style guide suggests one put defaults in hiera](https://puppet.com/docs/puppet/7.5/style_guide.html#parameter-defaults) i would even go so far as to call it a bug.  

further by puppet defaults in the class manifest it prevents you from doing things like

# common.yaml
```yaml
unbound::service_name: unbound
unbound::restart_cmd: "systemctl restart %l{ookup('unbound::service_name')}"
```

# OpenBSD.yaml (i.e. we dont need to redefine service_name)
```yaml
unbound::restart_cmd: "service restart %l{ookup('unbound::service_name')}"
```

bastelfreak:
I would also b e happy with puppet-strings could do that. I pinged in https://github.com/puppetlabs/puppet-strings/issues/250 again

bastelfreak (approved):

PR #270, opened 2021-04-27
[blacksmith] Bump version to 4.0.1-rc0

b4ldr:
prepare 4.0.0 release.  decided to do a major release as we drop puppet5 support otherwise its pretty minor

Also need to look at why the rake job bumped to 4.0.1-rc0 instead of 4.0.0-rc1


b4ldr:
This is not the right way to do this :S

PR #276, opened 2021-05-01
$tls_upstreami -> $tls_upstream typo

steadramon:
Noticed this typo - $tls_upstreami is used in init.pp but the template is looking for $tls_upstream


steadramon:
Noticed the label - this was a typo introduced in v4.0.0 - this worked in v3.0.0

kenyon:
Yeah, probably nobody else is using the typoed version of the parameter yet, so probably this should just be a bug fix.

PR #287, opened 2021-10-26
Require Unbound 1.6.6 or newer

### LINKED ISSUE #286: unbound_version not set on first run causing unexpected config file setting
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.25.0
- Ruby: 2.5.9
- Distribution: Ubuntu 20.04
- Module version: 4.0.1

## How to reproduce (e.g Puppet code you use)
```puppet
class { 'unbound':
  interface              => $interface_list,
  interface_automatic    => false,
  access                 => [ "${lookup('defaults::cidr')}", '127.0.0.0/8' ],
  do_not_query_localhost => false,
  val_permissive_mode    => true,
  ip_transparent         => true,
}
```
## What are you seeing
The initial run of the puppet agent fails to start unbound as the `ip-transparent` setting is not applied to the configuration file. As I am trying to bind to a failover IP address which may not exist at the time unbound is started unbound fails. The second run of puppet does add the setting and unbound starts correctly.

## What behaviour did you expect instead
The configuration file for unbound should contain all of the parameters specified such that unbound starts on the first run. 

## Output log
Can be provided if required.

## Any additional information you'd like to impart
The `unbound.conf.erb` template contains code to make parameters in the configuration specific to the version of unbound installed based on the fact `unbound_version`. This is not set on the first run of unbound as the package is not yet installed. All of the version dependent settings eg
```erb
    <%= print_config('so-reuseport', @so_reuseport, '1.4.22') -%>
    <%= print_config('ip-transparent', @ip_transparent, '1.5.4') -%>
    <%= print_config('ip-freebind', @ip_freebind, '1.5.9') -%>
```
Do not get applied as the `versioncmp` at the top of the template defaults to `0.a` and decides the version of unbound is lower than the minimum requirement for that setting. 

A potential fix could be to use a different variable in the template which is set after the package has been installed by running `unbound -V ` at that point. I'd be happy to have a go at creating a PR for this or an alternative solution.

b4ldr:
This PR updates the minimum supported version of unbund.  this allows
us to be a bit more generous when writing the first unbound.conf file
and fixes #286

I have picked unbound 1.6.6 as the minimum supported version.  this
seems to be the version available in centos 7 which i assume is the 
oldest version we support based on metatdata.json

Fixes: #286

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ A puppet module for the Unbound caching resolver.
 
 To use this module requires at least unbound 1.6.6.  Please also consult
 metadata.json to understand the minimum puppet version and any other module
-dependencies
+dependencies.
 
 ## Usage
 


bastelfreak:
@b4ldr can you take a look at the failing tests?

b4ldr:
> @b4ldr can you take a look at the failing tests?

done, at least im pretty sure the unbound issues are unrelated but will double check

ekohl (approved):
This feels like a major version bump to me. Perhaps we should first merge the other open PRs, get a release out and then do this major? Either way, I'm fine.

b4ldr:
> This feels like a major version bump to me. Perhaps we should first merge the other open PRs, get a release out and then do this major? 

SGTM

smortex (approved):
LGTM, my only concern is that nothing says '1.6.6' in the code :-). I am not familiar with this code base, and I let you judge if it makes sense to have the `print_config` function raise an error when passed a version older than the oldest supported version so that the CI can help locating settings that need an update when we change this version.

ekohl:
README is IMHO the best place.

metadata.json
```diff
@@ -94,6 +94,10 @@
     {
       "name": "puppet",
       "version_requirement": ">= 6.0.0 < 8.0.0"
+    },
+    {
+      "name": "unbound",
```
ekohl, line 99: I'm not sure I've ever seen a requirement different than Puppet. Is such a thing even documented? I'm not aware of any tool that reads this so I wonder what sense it makes.

README.md
```diff
@@ -23,8 +23,7 @@ A puppet module for the Unbound caching resolver.
 
 ## Requirements
 
-To use this module, you must be running at least Puppet 4.4.  If you are on an
-older version of Puppet, please use a 1.x version of this module.
+To use this module, you must be running at least Puppet 6 and unbound 1.6.6
```
ekohl, line 26: Perhaps it should refer to metadata.json for the correct Puppet version.

README.md
```diff
@@ -23,8 +23,7 @@ A puppet module for the Unbound caching resolver.
...
 
 To use this module, you must install the `puppetlabs/concat` and
```
ekohl, line 28: And this should IMHO refer to metadata.json too. Too easy to go out of sync.

metadata.json
b4ldr: no idea, i first just anted to see if CI would break in anyway, ill drop it from here and just leave t in the readme

README.md
b4ldr: ack sgtm will update

b4ldr:
updated

metadata.json
smortex: +1 for removing this.

smortex (commented):
Looks fine, just the dubious requirement…

metadata.json
b4ldr: done sorry forgot to add the change to the commit

README.md
```diff
@@ -23,18 +23,9 @@ A puppet module for the Unbound caching resolver.
...
-install`, or ensuring the following lines are present in your Puppetfile for an
-[R10k](https://github.com/puppetlabs/r10k) deployment.
-
-```Ruby
-mod 'concat', :git => 'git://github.com/puppetlabs/puppetlabs-concat.git'
-mod 'stdlib', :git => 'git://github.com/puppetlabs/puppetlabs-stdlib.git'
-```
+To use this module requires at least unbound 1.6.6.  Please also consult
+metadata.json to understand the minimum puppet version and any other module
+dependencies
```
ekohl, line 28:
```suggestion
dependencies.
```

ekohl (approved):

PR #290, opened 2021-11-24
unbound: don't allow empty string values

b4ldr:
Noticed this while reviewing #289

update `print_config` to ignore config values that have an empty string.
I believe this is an error for most parameters.

### DIFF
--- a/templates/unbound.conf.erb
+++ b/templates/unbound.conf.erb
@@ -5,7 +5,7 @@
     version = @unbound_version ? @unbound_version : '0.a'
     return version
   end
-  def print_config(name, value, version=false )
+  def print_config(name, value, version=false, allow_blank=true)
     if version and scope.call_function('versioncmp', [unbound_version, version]) < 0
        return
     end
@@ -15,7 +15,7 @@
       return "  #{name}: no\n"
     elsif not value
       return
-    elsif value.is_a?(String)
+    elsif value.is_a?(String) and (!value.empty? or allow_blank)
       return "  #{name}: \"#{value}\"\n"
     elsif value.is_a?(Integer)
       return "  #{name}: #{value}\n"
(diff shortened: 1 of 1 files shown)


bastelfreak:
@b4ldr can you please rebase against our latest master branch?

bastelfreak:
can you take a look at the failing tests? they are working on master.

b4ldr:
ack the ubuntu one seems unrelated, the centos ones look like they are to do with chroot , will try and get to them tomorrow

b4ldr:
Seems centos issues are pre-existing, considering centos is EOL is this something we still want to support?

templates/unbound.conf.erb
```diff
@@ -122,7 +122,7 @@ server:
     <%- end -%>
   <%- end -%>
 <%- end -%>
-<%= print_config('chroot', @chroot) -%>
+<%= print_config('chroot', @chroot, false, true) -%>
```
ekohl, line 125: I wonder if this wouldn't be clearer to have a separate if statement, just like on line 40.
```suggestion
<%= print_config('chroot', @chroot) -%>
```

ekohl (commented):
I wonder if silently ignoring them is the right thing. `allow_blank` implies that it refuses blank values, which in my mind also implies a hard error.

b4ldr:
> I wonder if silently ignoring them is the right thing. `allow_blank` implies that it refuses blank values, which in my mind also implies a hard error.

I agree in fact the more i think about it the more i think the entire approach of this patch is wrong.  If a value is allowed to have an empty string then we should just deal with that in the type signature, which as far as i can already looks pretty correct.  as such i think we should go back to the original patch and just create an if statement for that.

As such im going to close this down and i have created a new PR https://github.com/voxpupuli/puppet-unbound/pull/302 which is very similar to #289 with a few other minor tweaks

PR #295, opened 2022-03-04
Test ci

b4ldr:
have been debugging #290 and i have hit a few issues with centos
 
### centos7
seems to be using systemd as such it needs a privileged container.  Has this always been the case
 
### centos8
 "Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist"

Looks like its related to [EOL](https://www.centos.org/centos-linux-eol/).  creating this CR to see if things build by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Puppet powered DNS with Unbound
 
+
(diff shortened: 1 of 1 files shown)


b4ldr:
Seems centos is working now so will close this down

PR #299, opened 2022-06-03
Drop centos support

b4ldr:
CI is currently broken for all centos builds so we should drop it

See: https://github.com/voxpupuli/puppet-unbound/pull/295


b4ldr:
seems its no longer broken, ,so dropping this

PR #302, opened 2022-06-16
target-fetch-policy: is supposed to be a single string

b4ldr:
Slight modification of #289.  ensures the value is not nil, also update
the type signature to Optional[Array[Integer],1] to ensure that if an
array is passed its not empty


templates/unbound.conf.erb
```diff
@@ -147,7 +147,9 @@ server:
 <%= print_config('hide-version', @hide_version) -%>
 <%= print_config('version', @version) -%>
 <%= print_config('hide-trustanchor', @hide_trustanchor, '1.6.2') -%>
-<%= print_config('target-fetch-policy', @target_fetch_policy) -%>
+<%- unless @target_fetch_policy.nil? -%>
```
b4ldr, line 150: could also keep the type signature the same and check for empty? here, Happy to change i honestly can go either way on theses two styles depending on the day, but i think the current signiture is probably more specific, if a bit verbose.

PR #304, opened 2022-07-14
metadata.json: Fix source URL

bastelfreak:
the git protocol is deprecated, https should be used.


b4ldr:
i think it make senses to use https regardless but can you point to where the git protocol; got deprecated?

bastelfreak:
sure: https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

PR #326, opened 2024-01-18
Hiera and package management for Debian (made it for bookworm but could be usefull for any distro).

sfoutrel:
Hello, First time participant here.

Couldn't install on Debian 12 because 'unbound-anchor' was missing. That command was moved to another package. Though it would be usefull to modify class to allow more than one package to be installed.
Added some hiera for Debian 12 as well.

(No Issue declared).


b4ldr:
lgtm thanks

PR #332, opened 2024-02-08
Add systemd timer to update root.hints file

bastelfreak:
By default we download a root.hints file once. That's bad. it contains IP-addresses for all root DNS servers. Those addresses can change from time to time. We should update the file every now and then.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,6 +1,5 @@
-# Class: unbound
 #
-# Installs and configures Unbound, the caching DNS resolver from NLnet Labs
+# @summary Installs and configures Unbound, the caching DNS resolver from NLnet Labs
 #
 # @param hints_file
 #   File path to the root-hints. Set to 'builtin' to remove root-hint option from unbound.conf and use built-in hints.
@@ -8,6 +7,9 @@
 #   Contents of the root hints file, if it's not remotely fetched.
 # @param unbound_version
 #   the version of the installed unbound instance. defaults to the fact, but you can overwrite it. this reduces the initial puppet runs from two to one
+# @param update_root_hints
+#   If set to true (and hints_file isn't set to 'builtin') a systemd timer will be configured to update the root hints file every month
+#
 class unbound (
   Boolean                                       $manage_service                  = true,
   Integer[0,5]                                  $verbosity                       = 1,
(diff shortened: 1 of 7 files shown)


rwaffen (approved):

templates/roothints.service.epp
```diff
@@ -0,0 +1,9 @@
+<%- | Stdlib::Absolutepath $hints_file, Stdlib::HTTPSUrl $root_hints_url| -%>
+# THIS FILE IS MANAGED BY PUPPET
+# BASED ON https://wiki.archlinux.org/title/Unbound#Roothints_systemd_timer
+[Unit]
+Description=Update root hints for unbound
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/curl --output <%= $hints_file %> <%= $root_hints_url %>
```
evgeni, line 9: I guess this should use `unbound::fetch_client` and not `curl`?

evgeni (changes requested):

templates/roothints.service.epp
bastelfreak: I considered this, but! `unbound::fetch_client` doesn't have an absolute path and `ExecStart` requires this :( Should we use `ExecStart=/usr/bin/$unbound::fetch_client`? That's quite ugly, but it seems to be fine on CI.

evgeni: it's better than using `curl` here and `wget` there :)

also, I thought recent systemd is OK with relative paths here?

bastelfreak: okay I tested it locally and it works with a relative path. let's see what CI says.

evgeni: Seems EL7 is unhappy about that

bastelfreak: I hate old software so much

manifests/init.pp
```diff
@@ -316,6 +319,24 @@
       mode    => '0444',
       content => $hints_file_content,
     }
+    if $update_root_hints {
```
ekohl, line 322: Should it be able to ensure the timer is absent?

bastelfreak: implemented it

templates/roothints.service.epp
```diff
@@ -0,0 +1,9 @@
+<%- | Stdlib::Absolutepath $hints_file, Stdlib::HTTPSUrl $root_hints_url, String[1] $fetch_client | -%>
+# THIS FILE IS MANAGED BY PUPPET
+# BASED ON https://wiki.archlinux.org/title/Unbound#Roothints_systemd_timer
+[Unit]
+Description=Update root hints for unbound
+After=network.target
+
+[Service]
+ExecStart=<%= $fetch_client %> --output <%= $hints_file %> <%= $root_hints_url %>
```
evgeni, line 9: Fetch client already includes -O

kenyon:
I don't run DNS anymore, but when I did, I'd just use `/usr/share/dns/root.hints` that is installed by the [dns-root-data](https://tracker.debian.org/pkg/dns-root-data) package.

bastelfreak:
Also a nice idea, but that seems to be very Debian specific. Unbound also has hints compiled into it that can be used.

evgeni (approved):

PR #333, opened 2024-02-08
Drop EL7 support

bastelfreak:
It's dead this summer and it's blocking https://github.com/voxpupuli/puppet-unbound/pull/332


data/os/RedHat/6.yaml
evgeni, line 1: lol

evgeni (approved):

PR #345, opened 2024-06-11
Update local_zone_type.pp

and0x000:
update local zone types, see https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-local-zone


b4ldr:
thanks

PR #350, opened 2024-08-29
conf_d: merge the unbound_conf_d and conf_d params

### LINKED ISSUE #342: $conf_d and $unbound_conf_d  are not documented and unclear how they differ beyond their location
I'm trying to understand the difference between `$conf_d` and `$unbound_conf_d` in this module, when neither are used by default and are treated differently. As far as I can tell neither is used unless done so with an appropriate `$custom_server_conf` entry. In particular, `$conf_d` isn't purgeable and I'd like it to be based on how I understand it to be used.

Can these two parameters docs be clarified? If it turns out only one is needed, perhaps we can get rid the other?

b4ldr:
AFAICT the intention of theses two parameters is the same so lets merge them

fixes #342

### DIFF
--- a/data/os/Debian.yaml
+++ b/data/os/Debian.yaml
@@ -1,4 +1,5 @@
 ---
 unbound::pidfile: '/run/unbound.pid'
 unbound::runtime_dir: '/var/lib/unbound'
-unbound::purge_unbound_conf_d: true
+unbound::conf_d: "/etc/unbound/unbound.conf.d"
+unbound::purge_conf_d: true
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -207,7 +207,7 @@ The following parameters are available in the `unbound` class:
 * [`username`](#-unbound--username)
 * [`package_name`](#-unbound--package_name)
 * [`package_ensure`](#-unbound--package_ensure)
-* [`purge_unbound_conf_d`](#-unbound--purge_unbound_conf_d)
+* [`purge_conf_d`](#-unbound--purge_conf_d)
 * [`root_hints_url`](#-unbound--root_hints_url)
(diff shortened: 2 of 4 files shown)


data/os/Debian.yaml
```diff
@@ -1,4 +1,5 @@
 ---
 unbound::pidfile: '/run/unbound.pid'
 unbound::runtime_dir: '/var/lib/unbound'
-unbound::purge_unbound_conf_d: true
+unbound::conf_d: "/etc/unbound/unbound.conf.d"
```
b4ldr, line 4: I thought i could use the following here but `$confdir` doesn't appear to be in scope
```yaml
unbound::conf_d: "%{confdir}/unbound.conf.d"
```

bastelfreak: you can, but not with Puppet DSL interpolation, but with hiera interpolation:
```suggestion
unbound::conf_d: "%{lookup('confdir')}/unbound.conf.d"
```

b4ldr: If i do this then the spec test fails and when i look at the catalog the conf_d param is set to: ` File[/unbound.conf.d]` and we also see an untouched resource of `File[/unbound.conf.d]` in the coverage.  

Things i have tried

b4ldr: > i will accept this comment and let CI kick in case, just incase im going mad

bastelfreak: oh indeed this isn't going to work. `confdir` is a puppet variable. I assumed it's a hiera value. in Hiera you can only reference hiera keys (because all puppet code is evaluated afterwards). So you would need to move the data from init.pp to hiera.

b4ldr:
something has got messed up with this merge will resubmit

PR #356, opened 2025-01-17
Dnstap: Add new class to manage dnstap configueration

b4ldr:
Also add a new `unbound::print_config` function.  This is similar to the function already in `unbound.conf.erb` but written in puppet to allow us to migrate the unbound.conf template to epp in a further PR

### DIFF
--- a/functions/print_config.pp
+++ b/functions/print_config.pp
@@ -1,0 +1,21 @@
+# @summary Print a configuration value if it is defined and the version is supported
+# @param name the config item name
+# @param value the config item value
+# @param version the version when the config item was introduced
+# @return the config item as a string or an empty string if the version is not supported
+function unbound::print_config (
+  String[1]                                                     $name,
+  Optional[Variant[Boolean, Integer, String, Array[String, 1]]] $value   = undef,
+  Optional[String[1]]                                           $version = undef,
+) >> String {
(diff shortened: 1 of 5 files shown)


functions/print_config.pp
```diff
@@ -0,0 +1,21 @@
+# @summary Print a configuration value if it is defined and the version is supported
+# @param name the config item name
+# @param value the config item value
+# @param version the version when the config item was introduced
+# @return the config item as a string or an empty string if the version is not supported
+function unbound::print_config (
+  String                                                        $name,
```
bastelfreak, line 7: I think empty strings don't make sense here?
```suggestion
  String[1]                                                     $name,
```

functions/print_config.pp
```diff
@@ -0,0 +1,21 @@
...
+  Optional[Variant[Boolean, Integer, String, Array[String, 1]]] $value   = undef,
+  Optional[String[1]]                                           $version = undef,
+) >> Optional[String] {
```
bastelfreak, line 10: Is there a situation where `undef` is returned? I think it's always a string?

b4ldr: nice catch i think i started of retuning undef, but changed to '', fixed

functions/print_config.pp
b4ldr: done

bastelfreak (approved):

PR #357, opened 2025-01-17
migrate away from erb

b4ldr:
update the templating code to move away from erb and where possible use the print_config function

This PR is currently a WIP, couple of outstanding things
* we need to add a spec test for unbound::remote so before we merge this so we can ensure parity
* confirm it's ok to quote all string values
* spec tests are failing, but the error message suggests they shouldn't (hoping to get some help on this which is why im creating the pr early)

Also happy to split this up into separate PRs if needed


functions/clean_blank_lines.pp
```diff
@@ -0,0 +1,8 @@
+# @summary Remove blank lines from a string
+# @param content The content to remove blank lines from
+# @return The content with blank lines removed
+function unbound::clean_blank_lines (
```
bastelfreak, line 4: I think this is generic enough that it could also be part of extlib?

manifests/localzone.pp
```diff
@@ -26,15 +26,29 @@
 # @param template_name Use a custom template.
 #
 define unbound::localzone (
-  Unbound::Local_zone_type $type,
-  String $zone = $name,
-  $config_file = $unbound::config_file,
-  Array[Unbound::Resource_record_type] $local_data = [],
-  String $template_name = 'unbound/local_zone.erb'
+  Unbound::Local_zone_type             $type,
```
bastelfreak, line 29: I prefer the older style, with just a single whitespace. That makes the diff easier to read and makes git reverts / cherry-picks easier. But there's no consent in our style guide yet.

manifests/localzone.pp
```diff
@@ -26,15 +26,29 @@
...
+  String                               $zone          = $name,
+  Stdlib::Absolutepath                 $config_file   = $unbound::config_file,
+  String                               $template_name = 'unbound/local_zone.erb',
+  Array[Unbound::Resource_record_type] $local_data    = [],
 ) {
+  $records = $local_data.map |$record| {
```
bastelfreak, line 35: there's a lot going on here. Can you please add some comments about what's happening?

manifests/localzone.pp
b4ldr: I hear you but the rest of this module uses this style and I prefer consistency.  But consistency across viz would be great so let's try and get something in the style guide and then we can creat a puppet lint check and fix to get everything fixed?

manifests/localzone.pp
b4ldr: Fair comment will do

b4ldr: Updated let me know what's missing?

functions/clean_blank_lines.pp
b4ldr: https://github.com/voxpupuli/puppet-extlib/pull/239

b4ldr (commented):
if you have any idea on the spec test failure please say because im lost

PR #358, opened 2025-01-28
remote_spec: Add spec test for the unbound::remote class

b4ldr:


spec/classes/remote_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'unbound::remote' do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts.merge(concat_basedir: '/dne', unbound_version: '1.21.0') }
+      let(:params) do
```
bastelfreak, line 9: To distinguish between the local variable `facts` and the rspec variable `facts` can you please update it?
```suggestion
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) { os_facts.merge(concat_basedir: '/dne', unbound_version: '1.21.0') }
      let(:params) do
```

